PR #3, opened 2022-05-02
Main updates

ACC01ADE:
## Describe Changes

- Implementing ERC20 tokens
- Testing new standards enforcement process


contracts/Holograph.sol
```diff
@@ -101,7 +101,7 @@
 
 */
 
-pragma solidity 0.8.11;
+pragma solidity 0.8.13;
```
sogoiii: @thisisvitto you mentioned that 0.8.13 was failing. This should be 0.8.11 right? Lets isolate updating the solidity version into its own task.

.nvmrc
```diff
@@ -1 +1 @@
-v16.6.1
\ No newline at end of file
+v16.14.2
```
sogoiii: Since we are using asdf now we can remove .nvmrc

sogoiii (commented):
So when i run `_deploy.sh` it stops in `deploy/bridge_1...` with `Deployed MultichainERC721 Contract: 0x36648C73AeB836Cb93369D1B7202ADC1C219405b`

so it does not work for me locally.

I see that there are binaries to the different solidity version. That is not necessary.

ACC01ADE:
Can you try to run the latest version to see if it works for you now? @sogoiii

.nvmrc
ACC01ADE: Updated

contracts/Holograph.sol
ACC01ADE: For me it's important that minor version differences don't affect the overall code's functionality.
With latest work, was able to test between v0.8.11, v0.8.12, and v0.8.13 with success.
Now can leave it alone at v0.8.13 for the foreseeable future.

sogoiii (approved):
good progress. Merging to move code forward and make the next PRs easier.

PR #4, opened 2022-05-16
Feature/cxip 185 migrate to hardhat

ACC01ADE:
## Describe Changes

The bridge protocol has been migrated into a hardhat environment

- Deployment now runs through hardhat.
- Contracts are compiled from `src` dir into `contracts` dir every time `yarn compile` is run.
- `hardhat-deploy` has been forked into `hardhat-deploy-holographed` to allow for `HolographGenesis`-based deployment.
- Custom plugin was created for injecting code and automating code before compile.
- Prettier has been enabled into the project, and code can be formatted via `yarn prettier:fix` command.


ACC01ADE:
As of this moment, the repo is ready to merge with main branch. Hardhat is officially implemented in it's entirety. Also added first set of tests to ensure deployment succeeds and that deployed bytecode matches the compiled artifacts.

alexanderattar:
I noticed that the prettier doesn't seem to be working. I think there's errors in the .prettierrc. When I have a minute I can give it a closer look and fix

ACC01ADE:
> I noticed that the prettier doesn't seem to be working. I think there's errors in the .prettierrc. When I have a minute I can give it a closer look and fix

I was using the CLI version of prettier so the settings that editors use with the `prettier.` prepended in the variable names were not working for me. Ended up having to remove those in order for it to properly work. I can re-introduce those versions in as well so that this would work inside an editor and on command line simultaneously.

alexanderattar:
Aha gotcha! I think this should be good to go then @thisisvitto

sogoiii (approved):
@thisisvitto I was able to run everything from a fresh start. FYI i modified how the tests describe and its were organized. 

You are free to approve when you see this. I look froward to the other tests.

PR #13, opened 2022-08-23
adding job nonce to HolographBridge

ACC01ADE:
## Adding Job Nonce to Bridge Requests

We are adding job nonce to all bridge request to ensure that a unique jobPayloadHash is generated each time and makes it easier to catalog cross-chain requests individually.

- Adding _jobNonce to HolographBridge smart contract
- Implemented the use of nonce in all BridgeIn calls


hardhat.config.ts
```diff
@@ -167,11 +164,6 @@ const config: HardhatUserConfig = {
       chainId: networks.mumbai.chain,
       accounts: [MUMBAI_PRIVATE_KEY],
     },
-    fuji: {
```
alexanderattar: why was fuji removed here?

alexanderattar:
Changes generally make sense to me

alexanderattar (approved):
lg

hardhat.config.ts
ACC01ADE: oops, fixed it

PR #29, opened 2022-09-06
Fix typo and add testnet section

alexanderattar:


sogoiii (approved):
Good catch on on the branch name.

PR #36, opened 2022-09-15
Feature/operator pod logic

ACC01ADE:
## Describe Changes

- Change A
- Change B


contracts/Interfaces.sol
```diff
@@ -152,6 +152,8 @@ contract Interfaces is Admin, Initializable {
     _chainIdMap[ChainIdType.EVM][250][ChainIdType.HOLOGRAPH] = 5;
     // rinkeby
     _chainIdMap[ChainIdType.EVM][4][ChainIdType.HOLOGRAPH] = 4000000001;
+    // goerli
+    _chainIdMap[ChainIdType.EVM][5][ChainIdType.HOLOGRAPH] = 4000000011;
```
alexanderattar, line 156: Are these Holograph IDs arbitrary?

src/Interfaces.sol
```diff
@@ -155,6 +161,8 @@ contract Interfaces is Admin, Initializable {
     _chainIdMap[ChainIdType.HOLOGRAPH][7][ChainIdType.LAYERZERO] = 11;
     // rinkeby
     _chainIdMap[ChainIdType.HOLOGRAPH][4000000001][ChainIdType.LAYERZERO] = 10001;
+    // goerli
+    _chainIdMap[ChainIdType.HOLOGRAPH][4000000011][ChainIdType.LAYERZERO] = 10021;
```
alexanderattar, line 165: Same with this. Just wondering where these id numbers are coming from or if you have some kind of system for adding new ones

alexanderattar (approved):

alexanderattar (changes requested):
Realized this is still draft and combined with the pod logic branch. Merge conflicts and other things we need to plan before we can bring into develop

contracts/Interfaces.sol
ACC01ADE: There is some logic to the numbers, 4294967295 is max value for 2^32 which is uint32 (our highest chain id value).
We use 4294967295 and 4294967294 for localhost network ids.
We reserve all numbers 4000000001 and up for testnets.
Mainnets start from 1 up to 4000000000.
As we identify/use new testnets or mainnets, we assign them the next available Holograph ChainID. Goerli, being a testnet, was assigned 4000000011, since that was the next available number when we were adding it to the list.

src/Interfaces.sol
ACC01ADE: See previous question's answer.

ACC01ADE:
> Realized this is still draft and combined with the pod logic branch. Merge conflicts and other things we need to plan before we can bring into develop

We can just copy the commit that handles only the Goerli logic if anything.

contracts/Interfaces.sol
alexanderattar: aha okay cool. thanks for explaining 👏

contracts/Holograph.sol
```diff
@@ -122,142 +130,100 @@ contract Holograph is Admin, Initializable, IHolograph {
...
-   * @dev For example:
-   *                   1 = Ethereum mainnet
-   *                   2 = Binance Smart Chain mainnet
-   *                   3 = Avalanche mainnet
-   *                   4 = Polygon mainnet
-   *                   etc.
-   */
-  function getChainType() public view returns (uint32 chainType) {
-    // The slot hash has been precomputed for gas optimizaion
-    // bytes32 slot = bytes32(uint256(keccak256('eip1967.Holograph.Bridge.chainType')) - 1);
```
alexanderattar, line 151: Do you think it's not important to document what the slots were generated from in the comments still? Maybe it's worth putting one comment at the top of https://github.com/holographxyz/holograph-protocol/pull/36/files#diff-d7e82dd160b5facd64a9d67736cacc298beb65cc58c3ae9e7161f7ac9f686181R111 that explains the formula. Something like: 

`bytes32 slot = bytes32(uint256(keccak256('eip1967.Holograph.<contract-name>.<property-name>')) - 1);`

contracts/HolographBridge.sol
```diff
@@ -154,251 +153,209 @@ contract HolographBridge is Admin, Initializable, IHolographBridge {
...
-        ERC721Holograph.holographBridgeIn.selector,
-      "HOLOGRAPH: bridge in failed"
+      _registry().isHolographedContract(holographableContract) || address(_factory()) == holographableContract,
+      "HOLOGRAPH: not holographed"
     );
+    bytes4 selector = HolographableEnforcer(holographableContract).bridgeIn(fromChain, data);
+    require(selector == HolographableEnforcer.bridgeIn.selector, "HOLOGRAPH: bridge in failed");
+    if (hTokenValue > 0) {
+      // provide operator with hToken value for executing bridge job
+      require(
```
alexanderattar, line 185: Does this actually send out the value, or just check it?

contracts/HolographBridge.sol
```diff
@@ -154,251 +153,209 @@ contract HolographBridge is Admin, Initializable, IHolographBridge {
...
+      jobNonce + 1,
+      _holograph().getChainType(),
+      holographableContract,
+      _registry().getHToken(_holograph().getChainType()),
+      address(0),
+      0,
+      payload
     );
-    _factory().deployHolographableContract(config, signature, signer);
+    samplePayload = abi.encodePacked(encodedData, type(uint256).max, type(uint256).max);
```
alexanderattar, line 252: what is `samplePayload` used for?

contracts/HolographERC20.sol
```diff
@@ -183,15 +186,37 @@ contract HolographERC20 is Admin, Owner, Initializable, NonReentrant, EIP712, ER
...
    * @notice Initializes the collection.
    * @dev Special function to allow a one time initialisation on deployment. Also configures and deploys royalties.
    */
   function init(bytes memory data) external override returns (bytes4) {
     require(!_isInitialized(), "ERC20: already initialized");
+    IInitializable sourceContract;
     assembly {
-      sstore(0x5bf044ad06f1ee9ee2fcd7caf4e944279b877408c76c4515cca6b4e937b01334, 1)
-      sstore(0x89b583059fdb0b2e807359b64eba1a8a1e6d099210701fafe6dad5dd2cd64fb8, caller())
+      sstore(_reentrantSlot, 0x0000000000000000000000000000000000000000000000000000000000000001)
```
alexanderattar, line 217: is this for the jobNonce here?

contracts/HolographERC721.sol
```diff
@@ -491,68 +511,65 @@ contract HolographERC721 is Admin, Owner, ERC721Holograph, Initializable {
...
-    uint32 chain = _chain();
-    uint256 token;
-    for (uint256 i = 0; i < tokenIds.length; i++) {
-      require(!_burnedTokens[token], "ERC721: can't mint burned token");
-      token = uint256(bytes32(abi.encodePacked(chain, tokenIds[i])));
-      require(!_burnedTokens[token], "ERC721: can't mint burned token");
-      _mint(to, token);
-    }
-  }
+  //   function sourceMintBatch(address to, uint224[] calldata tokenIds) external onlySource {
```
alexanderattar, line 521: are these going to be reenabled?

contracts/HolographFactory.sol
```diff
@@ -133,25 +137,47 @@ contract HolographFactory is Admin, Initializable {
...
+
   /**
    * @dev A sample function of the deployment of bridgeable smart contracts.
    * @dev The used variables and formatting is not the final or decisive version, but the general idea is directly portrayed.
    * @notice In this function we have incorporated a secure storage function/extension. Keep in mind that this is not required or needed for bridgeable deployments to work. It is just a personal development choice.
    */
   function deployHolographableContract(
-    DeploymentConfig calldata config,
-    Verification calldata signature,
+    DeploymentConfig memory config,
```
alexanderattar, line 177: just curious why these were changed? is it so the function can be called internally now too?

contracts/Holographer.sol
```diff
@@ -112,6 +112,12 @@ import "./interface/IInitializable.sol";
  * @dev This contract is a binder. It puts together all the variables to make the underlying contracts functional and be bridgeable.
  */
 contract Holographer is Admin, Initializable {
+  bytes32 constant _originChainSlot = 0xd49ffd6af8249d6e6b5963d9d2b22c6db30ad594cb468453047a14e1c1bcde4d;
```
alexanderattar, line 115: Is there already a test that checks all these slot values are correct? I remember a while back there was a time when some typos were missed and some slots were incorrect. Just thinking of ways we could prevent something like that from accidentally happening in the future

contracts/PA1D.sol
```diff
@@ -270,7 +258,7 @@ contract PA1D is Admin, Owner, Initializable {
    * @return bp Royalty base points(percentage) for the royalty payouts of a specific token id.
    */
   function _getBp(uint256 tokenId) private view returns (uint256 bp) {
-    bytes32 slot = bytes32(uint256(keccak256(abi.encodePacked("eip1967.Holograph.PA1D.bp", tokenId))) - 1);
+    bytes32 slot = bytes32(uint256(keccak256(abi.encodePacked(_bpString, tokenId))) - 1);
```
alexanderattar, line 261: why are these slots not defined at the top of the contract like the others?

contracts/abstract/NonReentrant.sol
```diff
@@ -102,35 +102,26 @@
...
     assembly {
-      status := sload(
-        /* slot */
-        0x5bf044ad06f1ee9ee2fcd7caf4e944279b877408c76c4515cca6b4e937b01334
-      )
+      status := sload(_reentrantSlot)
     }
   }
 
   function setStatus(uint256 status) internal {
```
alexanderattar, line 122: This can only be 1 or 2 with 2 signaling that reentrancy has occurred before reverting?

contracts/enum/ChainIdType.sol
```diff
@@ -106,7 +106,5 @@ enum ChainIdType {
   EVM, //        1
   HOLOGRAPH, //  2
   LAYERZERO, //  3
-  SYNAPSE, //    4
-  CONNECTX, //   5
-  ABACUS //      6
+  ABACUS //      4
```
alexanderattar: Should this be updated to Hyperlane?

contracts/Holograph.sol
ACC01ADE: Sure, we can add it as commentary on top where storage slots are computed and stored as constants.

contracts/HolographBridge.sol
ACC01ADE: This is here for now to keep the code from reverting while HLG logic is being implemented.

contracts/HolographBridge.sol
ACC01ADE: This is for front-end to be able to run hypothetical bridge request and see what the payload will be, so that they can then run it on destination chain for an exact gas limit calculation.

contracts/HolographERC721.sol
ACC01ADE: Not at the moment. Once we got this iteration working, then we will enable/introduce batch minting and bridging.

contracts/HolographERC20.sol
ACC01ADE: This is initializing re-entrant slot to be non zero. We use 1 and 2 for re-entrancy checking. To make sure non-initialized contract doesn't get abused.

contracts/HolographFactory.sol
ACC01ADE: We enabled bridgeIn for HolographFactory, this resulted in the `deployHolographableContract` to be called internally, which required to temporarily switch to memory instead of calldata in order to be able to easily pass data over.

contracts/Holographer.sol
ACC01ADE: Not yet, it's on the todo list. The bridge will be responsible for doing a quick sanity check and make sure chain id did not change since last transaction. This is to prevent a hard fork and duplication of assets due to the fork.

contracts/PA1D.sol
ACC01ADE: These are dynamically generate values that rely on tokenId. Can't have them as a constant since it's for a per-token-basis.

contracts/abstract/NonReentrant.sol
ACC01ADE: yep

contracts/enum/ChainIdType.sol
ACC01ADE: yep

contracts/HolographBridge.sol
alexanderattar: gotcha

contracts/PA1D.sol
alexanderattar: aha missed that token id was being passed in

alexanderattar (approved):
Thanks for answering my questions. Let's get this into experimental

PR #39, opened 2022-10-01
Feature/holo 390 faucet smart contract

ackvf:
## Describe Changes

- Faucet contract


README.md
```diff
@@ -124,6 +124,38 @@ To manually run just the build task use `yarn run build`.
...
+
+   ```
+   remixd -s contracts
+   ```
+
+2. Run Ganache GUI and spin up a local blockchain (using workspace)
+3. Open Remix IDE
+   1. On the "**File explorer**" card change "Workspaces" -> "- connect to localhost -"
+   2. On the "**Deploy & Run Transactions**" card pick "Environment" -> "Ganache Provider"  
+   The localhost port is usually `7545` or `8545`
```
ACC01ADE, line 150: Ports are specifically 8545 (chain1) and 9545 (chain2).

README.md
```diff
@@ -124,6 +124,38 @@ To manually run just the build task use `yarn run build`.
...
+4. After making changes to the contracts in `/src/contracts` run `yarn dev:iterate` to update `/contracts`  
+   *(Watch script is coming later)*
+5. Development
+   1. In "**File Explorer**" card open each relevant contract as a tab
+   2. Enable "**Auto compile**" in "Solidity Compiler" card
+   3. Interact with the contracts on "Deploy & run transactions" tab  
+      To deploy a contract, first focus its tab in the IDE as this changes the "CONTRACT" dropdown.
```
ACC01ADE, line 157: You cannot deploy the protocol contracts via an IDE. It's too complex and requires running through HolographGenesis contract or through HolographFactory. An IDE is not configured for something like that.
IDE is only as good as using a block explorer to view transactions, or make direct web3 calls using contract ABI.

README.md
```diff
@@ -124,6 +124,38 @@ To manually run just the build task use `yarn run build`.
 
 _If you need the smart contracts ABI files for dApp integrations, use `yarn run abi` to get a complete list of all ABI's inside of the `abi` directory._
 
+### Development with live interaction & manual testing (supports Metamask)
```
alexanderattar, line 127: I do not think we should be instructing users to use Ganache / Remix. Currently we're using hardhat for deployment and testing

src/faucet/Faucet.sol
```diff
@@ -0,0 +1,90 @@
...
+    token = ERC20Holograph(tokenInstance_);
+    owner = contractOwner_;
+    _setInitialized();
+    return IInitializable.init.selector;
+  }
+
+  /// @notice Get tokens from faucet's own balance. Rate limited.
+  function requestTokens() external {
+    require(isAllowedToWithdraw(msg.sender), 'Come back later');
+    require(token.balanceOf(address(this)) >= faucetDripAmount, "Faucet is empty");
```
alexanderattar, line 40: does this mean the faucet can only have 100 tokens at a time?

src/faucet/Faucet.sol
```diff
@@ -0,0 +1,90 @@
...
+    require(token.balanceOf(address(this)) >= amountWei_, "Insufficient funds");
+    token.transfer(address_, amountWei_);
+  }
+
+  /// @notice Withdraw funds from the faucet.
+  function withdrawTokens(address receiver_) external onlyOwner {
+    token.transfer(receiver_, token.balanceOf(address(this)));
+  }
+
+  function withdrawTokens(address receiver_, uint256 amountWei_) external onlyOwner {
```
alexanderattar, line 62: Should these have different names? `withdrawAllTokens` and `withdrawTokens`?

src/faucet/Faucet.sol
```diff
@@ -0,0 +1,90 @@
...
+  /// @notice Get tokens from faucet's own balance. Rate limited.
+  function requestTokens() external {
+    require(isAllowedToWithdraw(msg.sender), 'Come back later');
+    require(token.balanceOf(address(this)) >= faucetDripAmount, "Faucet is empty");
+    lastAccessTime[msg.sender] = block.timestamp;
+    token.transfer(msg.sender, faucetDripAmount);
+  }
+
+  /// @notice Grant tokens to receiver from faucet's own balance. Not rate limited.
+  function grantTokens(address address_) external onlyOwner {
```
alexanderattar, line 46: This is a nit, but it's more typical for function variables to use the underscore at the front of the variable name in Solidity https://ethereum.stackexchange.com/questions/56443/what-does-mean-before-a-variable-name

alexanderattar (commented):
Tests just need to be finished

README.md
ackvf: Oh yeah, that was for a different way altogether, but I will update it to work with the existing two ganache instances instead!

README.md
ackvf: Yes. This was for a contract that has not yet been made holographable.

README.md
ackvf: It is very difficult to develop new contracts in this repo and to make the whole workflow working with deployment script and tests on first try.

The way I developed the Faucet contract was:

1) First develop the generic faucet contract using Remix and Ganache GUI.
2) Test and verify that the expected AC is met
3) After I had working Faucet contract that met all requirements I started writing test cases from the actual test runs I did by hand.
4) After that I wrote the deployment script


I can't imagine working in reverse order. There is so many pieces that need to go together an the Development experience in this repo just isn't great.

First I need a contract file in order to get hardhat artifacts, ABIs and Typechain types.
Then I need to write a deployment script and then test case before I get to testing the contract itself.

The workflow above gave me so much flexibility. Suddenly I could change a variable in the contract and immediately see it working in Remix and I could quickly iterate on the fine details.

Without this, I would always had to adjust tests first in order to see how it changed.

This repo is more suited for TDD. Sure I could work this way. I didn't in this instance and this is the flow.

How do you develop with this repo? A training session would perhaps come handy.

@alexanderattar @ACC01ADE

src/faucet/Faucet.sol
ackvf: It means that if faucet's balance is not larger or equal the drip amount, then it will revert.

In other words: If faucet balance is at least 100 tokens, it will not revert.

src/faucet/Faucet.sol
ackvf: Not necessary, solidity supports function overrides fine. Functions in solidity are identified by the whole signature.

src/faucet/Faucet.sol
ackvf: Interesting, I saw using underscore at the start of a variable name for private values and functions. Putting them at the end is my way of differentiating private variables from local arguments. ¯\\_(ツ)_/¯

ackvf: Yeah, you can see it in our repo too

```
contract CxipERC721 is ERC721H {
  /**
   * @dev Internal reference used for minting incremental token ids.
   */
  uint224 private _currentTokenId;

  /**
   * @dev Enum of type of token URI to use globally for the entire contract.
   */
  TokenUriType private _uriType;

  /**
   * @dev Enum mapping of type of token URI to use for specific tokenId.
   */
  mapping(uint256 => TokenUriType) private _tokenUriType;

  /**
   * @dev Mapping of IPFS URIs for tokenIds.
   */
  mapping(uint256 => mapping(TokenUriType => string)) private _tokenURIs;

  /**
   * @notice Get's the URI of the token.
   * @return string The URI.
   */
  function tokenURI(uint256 _tokenId) external view onlyHolographer returns (string memory) {
    TokenUriType uriType = _tokenUriType[_tokenId];
    if (uriType == TokenUriType.UNDEFINED) {
      uriType = _uriType;
    }
    return
      string(
        abi.encodePacked(
          IInterfaces(IHolograph(IHolographer(holographer()).getHolograph()).getInterfaces()).getUriPrepend(uriType),
          _tokenURIs[_tokenId][uriType]
        )
      );
  }

  function bridgeIn(
    uint32, /* _chainId*/
    address, /* _from*/
    address, /* _to*/
    uint256 _tokenId,
    bytes calldata _data
  ) external onlyHolographer returns (bool) {
    (TokenUriType uriType, string memory tokenUri) = abi.decode(_data, (TokenUriType, string));
    _tokenUriType[_tokenId] = uriType;
    _tokenURIs[_tokenId][uriType] = tokenUri;
    return true;
  }

  function bridgeOut(
    uint32, /* _chainId*/
    address, /* _from*/
    address, /* _to*/
    uint256 _tokenId
  ) external view onlyHolographer returns (bytes memory _data) {
    TokenUriType uriType = _tokenUriType[_tokenId];
    if (uriType == TokenUriType.UNDEFINED) {
      uriType = _uriType;
    }
    _data = abi.encode(uriType, _tokenURIs[_tokenId][uriType]);
  }
```

IMO this is very confusing and my approach is better.

ACC01ADE: I do like the idea of using underscores at end of variable name to define that it came from external data. Visually easy to understand this way where the data came from. Underscore in front for internal private variables, underscore at end to show user submitted var, no underscores means locally defined var (inside function). Could be interesting to try.

src/faucet/Faucet.sol
alexanderattar: Okay cool makes sense

src/faucet/Faucet.sol
alexanderattar: Yep i know that, but might make it easier to differentiate why there are two versions without having to think about it or inspect the parameters 🤷‍♂️

README.md
alexanderattar: I'm mainly suggesting we don't include these instructions in the README because it's going to be confusing for auditors when they look at the code. It's fine to keep these notes for yourself or we could create a notion doc for internal use. We can discuss when @ACC01ADE does a run through the codebase with the whole team the best process for this moving forward

src/faucet/Faucet.sol
alexanderattar: My opinion is to just stick with underscores in the front for now for consistency, but it's not a big deal either way

test/08_faucet_tests.ts
```diff
@@ -23,20 +24,21 @@ describe('Testing the Holograph Faucet', async () => {
...
   // Revert messages
-  const INITIALIZED = 'Faucet contract is already initialized'
+  const INITIALIZED = 'Faucet already initialized'
   const COME_BACK_LATER = 'Come back later'
   const NOT_AN_OWNER = 'Caller is not the owner'
 
   before(async function () {
     l1 = await setup()
-    SAMPLEERC20 = l1.sampleErc20.attach(l1.sampleErc20Holographer.address)
+    ERC20 = await l1.holographErc20.attach(l1.sampleErc20Holographer.address);
```
ackvf, line 28: `.attach` is not asynchronous

test/08_faucet_tests.ts
```diff
@@ -60,48 +62,46 @@ describe('Testing the Holograph Faucet', async () => {
 
     it('isAllowedToWithdraw(): User is allowed to withdraw for the first time', async function () {
       expect(
-        await FAUCET.functions.isAllowedToWithdraw(DEPLOYER)
+        await FAUCET.isAllowedToWithdraw(DEPLOYER)
```
ackvf: There is no reason to remove this.

test/08_faucet_tests.ts
```diff
@@ -60,48 +62,46 @@ describe('Testing the Holograph Faucet', async () => {
...
       ).to.be.false
     })
 
     it('requestTokens(): User cannot withdraw for the second time', async function () {
-      expect(
-        await FAUCET.functions.requestTokens()
-      ).to.be.revertedWith(COME_BACK_LATER)
+      await expect(FAUCET.requestTokens()).to.be.revertedWith(COME_BACK_LATER)
```
ackvf: This is wrong. expect and its properties are not asynchronous (except `.resolves` an `.rejects`) https://jestjs.io/docs/asynchronous 


Correct would be
```
const result = await FAUCET.requestTokens()
expect(result).to.be.revertedWith(COME_BACK_LATER)
```

which is analogous to
```
expect(await FAUCET.requestTokens()).to.be.revertedWith(COME_BACK_LATER)
```

test/08_faucet_tests.ts
```diff
@@ -154,51 +148,49 @@ describe('Testing the Holograph Faucet', async () => {
...
       expect(
-        await SAMPLEERC20.functions.balanceOf(USER_A)
+        await ERC20.balanceOf(USER_A)
       ).to.equal(DEFAULT_DRIP_AMOUNT.mul(factor))
     })
 
     it('setWithdrawAmount(): Owner adjusts Withdraw Amount back to 100 eth', async function () {
       expect(
-        await FAUCET.functions.setWithdrawAmount(DEFAULT_DRIP_AMOUNT)
+        await FAUCET.setWithdrawAmount(DEFAULT_DRIP_AMOUNT)
```
ackvf: Removing all those `.functions` was unnecessary.

ackvf (commented):
@ACC01ADE Please see review comments

deploy/14_faucet.ts
```diff
@@ -1,82 +1,78 @@
...
+
+  const accounts = await hre.ethers.getSigners();
+  const deployer: SignerWithAddress = accounts[0];
+
+  const network = networks[hre.networkName];
+
+  const salt = hre.deploymentSalt;
+
+  const holographRegistryProxy = await hre.ethers.getContract('HolographRegistryProxy');
+  const holographRegistry = ((await hre.ethers.getContract('HolographRegistry')) as Contract).attach(
```
ackvf, line 27: Is there a reason to remove the correct typings for those contracts?
```
const holographRegistry = await hre.ethers.getContract<HolographRegistry>('HolographRegistry')
```

README.md
ackvf: OK, Notion sounds good

src/faucet/Faucet.sol
ackvf: renamed to `withdrawAllTokens`

README.md
ackvf: OK, Notion sounds good

src/faucet/Faucet.sol
ackvf: renamed to `withdrawAllTokens`

test/08_faucet_tests.ts
ackvf: Actually sorry, revertedWith is asynchronous! But `expect(await ...)` still requires to await async code inside.

test/08_faucet_tests.ts
ACC01ADE: Yep you're right, was doing copy paste and didn't pay it much attention. Will clean up the code from this in next iteration.

deploy/14_faucet.ts
ACC01ADE: It was a copy paste job from another deploy script. You can try to use correct typings. If it works for you, then can do the same for other deploy scripts in next iteration.

package.json
```diff
@@ -16,6 +16,7 @@
     "typechain-types"
   ],
   "scripts": {
+    "dev:iterate": "yarn compile & yarn typechain",
```
ACC01ADE, line 19: Running `yarn run clean-compile` accomplishes the same.

src/faucet/Faucet.sol
ACC01ADE: Let's leave it as it is for now and agree on a standard format later when we have more time to give the code a proper facelift?

README.md
ACC01ADE: We'll be working on more thorough documentation for devs at a later time, so this detailed explanations can be provided in the Notion doc as @alexanderattar suggested. I do like the idea of keeping the current README as minimal and to the point as possible for just cloning, building, running the thing and nothing else. We can provide a link at the top later that sends developers to a dedicated Notion section that goes into more details about the contracts, building, and interacting with them.

test/08_faucet_tests.ts
ACC01ADE: I removed the `functions` portion for code consistency, since all other tests do not have that in them at this point. I understand why you have them there, to visually better understand that it is being called from a `Contract` instance. Whether you use it or not in these tests is entirely up to you. It does not affect the functionality in any way if they are or are not there.

test/08_faucet_tests.ts
ACC01ADE: Await inside of the `expect` function is optional since it automatically handles promises. I do like the idea of still including it though, cuz it helps to better know how it's being run. And yes, `revertedWith` is asynchronous. As for breaking down the call and saving it as a `result` before using it in the `expect` call, that's your personal code preference. Feel free to do that, but please don't call it the "correct way", that is opinion.

test/08_faucet_tests.ts
ACC01ADE: Responded to this in same comment above ^^^
https://github.com/holographxyz/holograph-protocol/pull/39#discussion_r987922145

alexanderattar (approved):
Looks good

ACC01ADE:
Let's merge the PR #40 first and see if it creates conflicts to this branch before trying to merge it into experimental.

test/08_faucet_tests.ts
ackvf: BTW, there is another benefit of using `functions`, better intellisense in IDE as it displays only the functions without other Ethers Contract instance methods. However, I need to figure out why it gives different result than calling the functions directly.

@ACC01ADE @alexanderattar

test/08_faucet_tests.ts
ackvf: Interesting, from the docs and googling I understood that when you pass a promise in `expect()`, you need to use `resolves` or `rejects` ¯\\_(ツ)_/¯ and tbh I vaguely remember that when I didn't use await in expect, the tests would pass before it had chance to resolve IIRC.

ackvf: Oh wait a minute. We are not using Jest, but Chai?? That explains a lot,

PR #40, opened 2022-10-01
Feature/operator tests

ACC01ADE:
## Describe Changes

Improving operator pod logic with extensive hardhat tests, to ensure everything is working properly.

- Merged experimental branch in for most up to date code
- Enabled cross-chain smart contract deployments
- Updated beaming tests to work with new version of bridge request
- Provided on-chain functions to extract future payload and job params


contracts/Holograph.sol
```diff
@@ -107,130 +107,243 @@ import "./abstract/Initializable.sol";
...
       sstore(_registrySlot, registry)
       sstore(_treasurySlot, treasury)
+      sstore(_utilityTokenSlot, utilityToken)
     }
     _setInitialized();
     return IInitializable.init.selector;
   }
 
-  function getChainType() external view returns (uint32 chainType) {
+  /**
```
alexanderattar, line 164: Love all the new comments!

ACC01ADE: It's still a work in progress, starting little by little to document all the things.

contracts/HolographBridge.sol
```diff
@@ -201,95 +198,100 @@ contract HolographBridge is Admin, Initializable, IHolographBridge {
...
+      _registry().getHToken(_holograph().getHolographChainId()),
       address(0),
       0,
+      true,
       payload
     );
     _operator().send{value: msg.value}(gasLimit, gasPrice, toChain, msg.sender, encodedData);
   }
 
+  function revertedBridgeOutRequest(
```
alexanderattar, line 217: Is this just used for testing?

ACC01ADE: This is used for being able to properly get Bridge Out Request payload bytes without actually triggering a transaction. Running it as view or pure is impossible due to state changes being made. So we apply a simple trick where we use this function to force revert the internal call but still return the expected results. We then can get this by calling the getBridgeOutRequestPayload as a staticCall and retrieve the payload as it would be if the transaction actually happened, without actually executing the transaction.

contracts/HolographInterfaces.sol
```diff
@@ -457,7 +265,18 @@ contract Interfaces is Admin, Initializable {
...
+   * @dev This function is called by the deployer/factory when creating a contract.
+   */
+  function init(bytes memory data) external override returns (bytes4) {
+    require(!_isInitialized(), "HOLOGRAPH: already initialized");
+    address contractAdmin = abi.decode(data, (address));
+    assembly {
+      sstore(_adminSlot, contractAdmin)
+    }
     _setInitialized();
     return IInitializable.init.selector;
```
alexanderattar, line 281: I've noticed this pattern a lot in the holograph contracts. just curious why you return selector here? for use to confirm after contract is initialized

ACC01ADE: Yep, to ensure that the actual function was called and succeeded. Because there is no revert or error if a function is called on a non-deployed/existing contract. This way whenever we call a connected contract, we check for a function selector return to guarantee that the underlying logic exists and was executed successfully. A safeguard against improperly deployed or destroyed contracts since we allow for custom contracts to be implemented by developers.

contracts/HolographOperator.sol
```diff
@@ -186,20 +183,23 @@ contract HolographOperator is Admin, Initializable, IHolographOperator {
...
     require(msg.sender == _bridge(), "HOLOGRAPH: bridge only call");
     _;
   }
 
+  /**
+   * @dev Allow calls only from LayerZero endpoint contract.
+   */
   modifier onlyLZ() {
     assembly {
       // check if lzEndpoint
```
alexanderattar, line 205: It could be nice to have the hashes in the `mstore` be constants or add comments to some of these assembly blocks that are doing more than setting/getting a variable. OpenSea's SeaPort codebase does a really nice job of this: https://github.com/ProjectOpenSea/seaport/blob/171f2cd7faf13b2bf0455851499f1981274977f7/contracts/lib/GettersAndDerivers.sol

contracts/HolographOperator.sol
```diff
@@ -294,7 +297,9 @@ contract HolographOperator is Admin, Initializable, IHolographOperator {
       // decrease pod size to accomodate popped operator
       _operatorTempStorage[_operatorTempStorageCounter] = _operatorPods[pod][operatorIndex];
       _popOperator(pod, operatorIndex);
-      podSize--;
+      if (podSize > 1) {
```
alexanderattar, line 300: This to keep it as just one pod for now?

contracts/HolographOperator.sol
```diff
@@ -431,65 +436,99 @@ contract HolographOperator is Admin, Initializable, IHolographOperator {
...
+    uint256 pod
+  ) external {
+    require(_bondedOperators[operator] == 0, "HOLOGRAPH: operator is bonded");
+    unchecked {
+      uint256 current = _getCurrentBondAmount(pod - 1);
+      require(current <= amount, "HOLOGRAPH: bond amount too small");
+      // subtract difference and only keep bond amount
+      if (_operatorPods.length < pod) {
+        for (uint256 i = _operatorPods.length; i <= pod; i++) {
+          _operatorPods.push([address(0)]);
```
alexanderattar, line 505: Why pushing zero address here?

contracts/HolographOperator.sol
```diff
@@ -614,59 +662,13 @@ contract HolographOperator is Admin, Initializable, IHolographOperator {
...
-    _bondedAmounts[operator] = 0;
-  }
-
-  function bondUtilityToken(
-    address operator,
-    uint256 amount,
-    uint256 pod
-  ) external {
-    require(_bondedOperators[operator] == 0, "HOLOGRAPH: operator is bonded");
+  function _RBH(
```
alexanderattar, line 665: RBH is random block hash?

contracts/HolographRegistry.sol
```diff
@@ -105,21 +105,22 @@ import "./abstract/Admin.sol";
...
  * @dev This way it can be super easy to upgrade/update the source code once, and have all smart contracts automatically updated.
  */
-contract HolographRegistry is Admin, Initializable {
+contract HolographRegistry is Admin, Initializable, IHolographRegistry {
   bytes32 constant _holographSlot = 0xb4107f746e9496e8452accc7de63d1c5e14c19f510932daa04077cd49e8bd77a;
   bytes32 constant _utilityTokenSlot = 0xbf76518d46db472b71aa7677a0908b8016f3dee568415ffa24055f9a670f9c37;
 
   /**
-   * @dev A list of smart contracts that are guaranteed secure and holographable.
+   * @dev Array of all Holographable contracts that were ever deployed on this EVM chain
```
alexanderattar, line 121: This is cool

contracts/HolographRegistry.sol
```diff
@@ -138,22 +139,23 @@ contract HolographRegistry is Admin, Initializable {
...
    */
-  mapping(uint32 => address) private _hTokens;
+  mapping(address => bool) private _holographedContracts;
 
   /**
-   * @dev Array of all Holographable contracts that were ever deployed on this EVM chain
+   * @dev Mapping of all hTokens available for the different EVM chains
    */
-  address[] private _holographableContracts;
+  mapping(uint32 => address) private _hTokens;
```
alexanderattar, line 149: This would be a mapping per each deployment on each chain though. Wondering how this will be kept in sync from a global all chains perspective 🤔

contracts/HolographOperator.sol
ACC01ADE: This is to prevent an error for the random math if there is only one actively bonded operator in the pod, which at this point would be a zero address, allowing anyone to operate. Otherwise we would have the `_RBH` command fail since a zero is used in the math calculations.

alexanderattar (approved):
Tests passing ✅

contracts/HolographOperator.sol
ACC01ADE: I will provide descriptions and explanations inside of assembly functions to explain what is happening. This is gonna happen in the next iteration when I'll be commenting and documenting all the functions.

contracts/HolographOperator.sol
ACC01ADE: A pod cannot be empty, this way it allows for a zero address to be selected if the pod is selected. Zero address allows anyone to operate and is kind of like a winning the lotto scenario for operators.

contracts/HolographOperator.sol
ACC01ADE: Yes, I wrote it shorthand at that point due to time constraints, will rename it to make more clear in next iteration.

contracts/HolographRegistry.sol
ACC01ADE: Wanted to use a bit more gas on deployment but have a way to use the blockchain to store a historical array of all holographable contract ever deployed on that chain. Used with the `getHolographableContracts` paginated function allows us to retrieve them at any given point in time.

contracts/HolographRegistry.sol
ACC01ADE: Every time we deploy a new chain, we will use the hToken deploy code on already existing chains to update the list of supported hTokens. This way we will continuously keep updating the mappings on each chain every time a new chain is activated.

contracts/HolographInterfaces.sol
alexanderattar: nice thanks for explaining

contracts/HolographOperator.sol
alexanderattar: Nice sounds good

contracts/HolographOperator.sol
alexanderattar: ahh

contracts/HolographOperator.sol
alexanderattar: ahh okay that's cool

contracts/HolographOperator.sol
alexanderattar: cool

contracts/HolographRegistry.sol
alexanderattar: 👏

PR #50, opened 2022-10-11
HOLO-533: Add registry tests

sogoiii:
## Describe Changes

- Add tests for HolographRegistry contract
_note:_ I did not manage to implement the "inherited contract calls"

Screen Shot 2022-10-12 at 22 53 58
Screen Shot 2022-10-12 at 22 54 08


ACC01ADE:
@sogoiii 
```
  describe('init()', async function () {
    it.only('should successfully be initialized once', async function() {
      const initCode = generateInitCode(['address', 'bytes32[]'], [deployer.address, []]);
      await expect(holographRegistry.connect(deployer).init(initCode)).to.not.be.reverted;
    }); // Validate hardcoded values are correct
    it('should fail if already initialized');
    it('Should allow external contract to call fn');
    it('should fail to allow inherited contract to call fn');
  });
```

feliperodriguess (approved):
🤝

sogoiii:
👍

PR #56, opened 2022-10-12
Feature/HOLO 534 holograph treasury tests

BravoNatalie:
## Describe Changes

- Add tests for HolographTreasury
- Added MockExternalCall contract for testing external calls

image


test/16_holograph_treasury_tests.ts
```diff
@@ -1,80 +1,257 @@
...
+
+  const initPayload = ethers.utils.defaultAbiCoder.encode(
+    ['address', 'address', 'address', 'address'],
+    [bridgeMock, holographMock, operatorMock, registryMock]
+  );
+
+  beforeEach(async () => {
+    [deployer, commonUser] = await ethers.getSigners();
+
+    const holographTreasuryFactory = await ethers.getContractFactory('HolographTreasury');
```
ackvf, line 32: try this
```
    const holographTreasuryFactory = await ethers.getContractFactory<HolographTreasury__factory>('HolographTreasury');
    holographTreasury = await holographTreasuryFactory.deploy();
    await holographTreasury.deployed();
```

same for external call

ackvf (changes requested):

test/16_holograph_treasury_tests.ts
ackvf: No need to use type assertions.

sogoiii:
@ackvf i accidentally deleted your comment, thinking i was deleting my reply. :( sorry.

ackvf:
@sogoiii LOOOOL I didn't know you can do that 😂

test/16_holograph_treasury_tests.ts
```diff
@@ -1,80 +1,257 @@
...
+    it('should fail to allow inherited contract to call fn', async () => {});
+  });
+
+  describe('After initialized', () => {
+    beforeEach(async () => {
+      const tx = await holographTreasury.connect(deployer).init(initPayload);
+      await tx.wait();
+    });
+
+    function testIfIsPrivate(fnAbi: string, fnName: string, args: any[] = []) {
```
sogoiii, line 89: Oh this is nice!

test/16_holograph_treasury_tests.ts
```diff
@@ -1,80 +1,257 @@
...
+            to: holographTreasury.address,
+            data: encodedFunctionData,
+          })
+        ).to.be.reverted;
+      });
+    }
+
+    describe(`_bridge()`, async function () {
+      // it('should successfully be initialized once')
+      testIfIsPrivate('function _bridge() view returns (address bridge)', '_bridge');
```
feliperodriguess, line 105: This seems to be passing for public functions as well

test/16_holograph_treasury_tests.ts
```diff
@@ -1,80 +1,257 @@
...
+import { generateInitCode, zeroAddress } from '../scripts/utils/helpers';
+import { HolographTreasury, MockExternalCall } from '../typechain-types';
+
 describe('Holograph Treasury Contract', async function () {
+  let holographTreasury: HolographTreasury;
+  let mockExternalCall: MockExternalCall;
+  let deployer: SignerWithAddress;
+  let commonUser: SignerWithAddress;
+
+  const ONLY_ADMIN_ERROR_MSG = 'HOLOGRAPH: admin only function';
```
sogoiii: maybe add this to `/utils/error_constants.ts`

test/16_holograph_treasury_tests.ts
ackvf: Damn, how do you propose to solve that? How can we test that?

sogoiii: What do you mean?

feliperodriguess: I used this same helper function for my test and seems to be passing when I provide a public one

sogoiii: @ackvf @feliperodriguess  I have my snippet to test if the function is private. Granted it is a bit silly.

```
    it('is private function', async function () {
      expect(typeof holographOperator.connect(deployer)._randomBlockHash).to.equal('undefined')
      expect(holographOperator.connect(deployer)).to.not.have.keys('_randomBlockHash')
    });
  ```

Comment (ackvf) [reply test/16_holograph_treasury_tests.ts:105]: @sogoiii I wonder how this works? Does this test on the javascript level or does it do an actual blockchain call?

Comment (sogoiii) [reply test/16_holograph_treasury_tests.ts:105]: @ackvf it works by a negative. If hardhat reads the ABI and it creates the function, then we know that the function is no longer private. Since it is private, hardhat wont make the function and add it to the contract object.

Comment (BravoNatalie) [reply test/16_holograph_treasury_tests.ts:105]: > I used this same helper function for my test and seems to be passing when I provide a public one

You can't put this function inside another `it`, otherwise it not actually executed and the test seems to pass

Comment (feliperodriguess) [reply test/16_holograph_treasury_tests.ts:105]: yep, you're right. All good @BravoNatalie ✅

File: test/utils/error_constants.ts [hunk 5]
```diff
@@ -0,0 +1,2 @@
+export const ONLY_ADMIN_ERROR_MSG = 'HOLOGRAPH: admin only function';
```
sogoiii, line 1: 👍

sogoiii (changes requested):
Just fix the conflicts and we are good to go.

sogoiii (approved):
👍

PR #64, opened 2022-10-14
HOLO-544: Add abstract ERC721H tests

feliperodriguess:
## Describe Changes

- Added tests for `contracts/abstract/ERC721H.sol` on `19_ERC721H_tests.ts`
- ps: There are methods of this contract file that are not within the test file because part of them are `internal` and not really testable functions


sogoiii (approved):
👍

PR #81, opened 2022-11-09
Improvement/ HOLO 595 Enforce prettier / lint run on protocol

BravoNatalie:
## Describe Changes

- added `workflows/enforce_prettier.yml`
- added `workflowsqenforce_eslint.yml`
- added eslint configuration


.github/workflows/enforce_prettier.yml
```diff
@@ -0,0 +1,34 @@
...
+          fetch-depth: 0
+          ref: ${{ github.head_ref }}
+          # Make sure the value of GITHUB_TOKEN will not be persisted in repo's config
+          persist-credentials: false
+
+      - name: Prettify code
+        uses: creyD/prettier_action@v4.2
+        with:
+          prettier_options: --check deploy/**/*.ts test/**/*.ts scripts/**/*.ts
+          dry: True
```
alexanderattar, line 34: does this mean?

alexanderattar (approved):
Cool this will be nice to have

alexanderattar (changes requested):
image

There's a lot of options in the prettier config that are being ignored. I think we can remove those. Also this is currently running the prettier action in as a dry-run. I believe we need to set `dry` to false so it actually runs the changes

alexanderattar:
On further inspection of the previously run action, I noticed some issues. See above ☝️

.github/workflows/enforce_prettier.yml
BravoNatalie: it runs the prettier check on the .ts files under this folders

BravoNatalie:
> image
> 
> There's a lot of options in the prettier config that are being ignored. I think we can remove those. Also this is currently running the prettier action in as a dry-run. I believe we need to set `dry` to false so it actually runs the changes

This's happening because the prettier file has these rules, I haven't removed them because I thought the file was created like that on purpose. But if you agree I can remove them.

alexanderattar:
Yep those rules are my bad. I removed them and pushed the change

alexanderattar (approved):

sogoiii (approved):
👍

PR #82, opened 2022-11-11
Improvement/holo 603 gas optimization combine address mappings

alexanderattar:
## Describe Changes

>G-01 Multiple address mappings can be combined into a single mapping of an address to a struct
Multiple address mappings can be combined into a single mapping of an address to a struct, where appropriate saves a storage slot for the mapping. Depending on the circumstances and sizes of types, can avoid a Gsset (20000 gas) per mapping combined. Reads and subsequent writes can also be cheaper when a function requires both values and they both fit in the same storage slot.

- Updated - HolographERC20.sol
- Updated - HolographERC721.sol 
- Updated - HolographOperator.sol


alexanderattar:
@ACC01ADE I realize the gas savings from this might not be much as the values are not used in the same function and don't fit into the same slot, but this was my way of picking something easy up to get more comfortable updating the contracts before getting into more complicated stuff. If we feel that the potential gas savings don't justify the introduction of the new struct types / ergonomic changes of accessing the values, it's fine to not integrate these changes. lmk what you think

alexanderattar:
You can see some calls that have some gas savings in the screenshots below

alexanderattar:
HolographERC721 (Before)
image


HolographERC721 (After)
image

alexanderattar:
HolographERC20 (Before)

image


HolographERC20 (After)
image

alexanderattar:
HolographERC20 (Before)
image

HolographERC20 (After)
image


HolographERC721 (Before)
image

HolographERC721 (After)
image

alexanderattar:
HolographOperator (Before)
image

HolographOperator (After)
image

PR #93, opened 2022-11-30
Feature/holo 612 royalty smart contract improvements

alexanderattar:
### Royalty contract improvements based on feedback from C4 audit:

- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/33
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/164
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/180
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/186
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/377
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/384
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/451
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/454
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/455
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/456
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/476
- [x] https://github.com/code-423n4/2022-10-holograph-findings/issues/490

### DIFF
--- a/contracts/enforcer/HolographRoyalties.sol
+++ b/contracts/enforcer/HolographRoyalties.sol
@@ -398,7 +398,7 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     uint256 length = addresses.length;
     // accommodating the 2300 gas stipend
     // adding 1x for each item in array to accomodate rounding errors
-    uint256 gasCost = (23300 * length) + length;
+    uint256 gasCost = (2300 * length) + length;
     uint256 balance = address(this).balance;
     require(balance - gasCost > 10000, "ROYALTIES: Not enough ETH");
     balance = balance - gasCost;
@@ -406,7 +406,9 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     // uint256 sent;
     for (uint256 i = 0; i < length; i++) {
       sending = ((bps[i] * balance) / 10000);
-      addresses[i].transfer(sending);
+      // addresses[i].transfer(sending);
+      (bool success, ) = addresses[i].call{value: sending}("");
+      require(success, "ROYALTIES: Transfer failed");
       // sent = sent + sending;
     }
   }
--- a/contracts/enforcer/HolographRoyalties.sol
+++ b/contracts/enforcer/HolographRoyalties.sol
@@ -396,20 +396,13 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     address payable[] memory addresses = _getPayoutAddresses();
     uint256[] memory bps = _getPayoutBps();
     uint256 length = addresses.length;
-    // accommodating the 2300 gas stipend
-    // adding 1x for each item in array to accomodate rounding errors
-    uint256 gasCost = (2300 * length) + length;
     uint256 balance = address(this).balance;
-    require(balance - gasCost > 10000, "ROYALTIES: Not enough ETH");
-    balance = balance - gasCost;
+    require(balance > 10000, "ROYALTIES: Not enough ETH");
     uint256 sending;
-    // uint256 sent;
     for (uint256 i = 0; i < length; i++) {
       sending = ((bps[i] * balance) / 10000);
-      // addresses[i].transfer(sending);
       (bool success, ) = addresses[i].call{value: sending}("");
       require(success, "ROYALTIES: Transfer failed");
-      // sent = sent + sending;
     }
   }
 
@@ -425,11 +418,16 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     uint256 balance = erc20.balanceOf(address(this));
     require(balance > 10000, "ROYALTIES: Not enough tokens");
     uint256 sending;
-    //uint256 sent;
     for (uint256 i = 0; i < length; i++) {
       sending = ((bps[i] * balance) / 10000);
-      require(erc20.transfer(addresses[i], sending), "ROYALTIES: ERC20 transfer failed");
-      // sent = sent + sending;
+      // Some tokens revert when transferring a zero value amount this check ensures if one recipient's
+      // amount is zero, the transfer will still succeed for the other recipients.
+      if (sending > 0) {
+        // TODO: Need to use safeTransferFrom to support non-compliant tokens such as USDT, BNB, etc.
+        // See: https://github.com/code-423n4/2022-10-holograph-findings/issues/456
+        (bool success, ) = addresses[i].call{value: sending}("");
+        require(success, "ROYALTIES: ERC20 transfer failed");
+      }
     }
   }
 
@@ -448,11 +446,16 @@ contract HolographRoyalties is Admin, Owner, Initializable {
       erc20 = ERC20(tokenAddresses[t]);
       balance = erc20.balanceOf(address(this));
       require(balance > 10000, "ROYALTIES: Not enough tokens");
-      // uint256 sent;
       for (uint256 i = 0; i < addresses.length; i++) {
         sending = ((bps[i] * balance) / 10000);
-        require(erc20.transfer(addresses[i], sending), "ROYALTIES: ERC20 transfer failed");
-        // sent = sent + sending;
+        // Some tokens revert when transferring a zero value amount this check ensures if one recipient's
+        // amount is zero, the transfer will still succeed for the other recipients.
+        if (sending > 0) {
+          // TODO: Need to use safeTransferFrom to support non-compliant tokens such as USDT, BNB, etc.
+          // See: https://github.com/code-423n4/2022-10-holograph-findings/issues/456
+          (bool success, ) = addresses[i].call{value: sending}("");
+          require(success, "ROYALTIES: ERC20 transfer failed");
+        }
       }
     }
   }
@@ -487,6 +490,8 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     require(addresses.length == bps.length, "ROYALTIES: missmatched lenghts");
     uint256 totalBp;
     for (uint256 i = 0; i < addresses.length; i++) {
+      require(addresses[i] != address(0), "ROYALTIES: payee is zero address");
+      require(bps[i] > 0, "ROYALTIES: bp is zero");
       totalBp = totalBp + bps[i];
     }
     require(totalBp == 10000, "ROYALTIES: bps must equal 10000");
@@ -546,6 +551,7 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     address payable receiver,
     uint256 bp
   ) public onlyOwner {
+    require(bp <= 10_000, "ROYALTIES: cannot be over 100%");
     if (tokenId == 0) {
       _setDefaultReceiver(receiver);
       _setDefaultBp(bp);
@@ -682,9 +688,9 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     bidShares.prevOwner.value = 0;
     bidShares.owner.value = 0;
     if (_getReceiver(tokenId) == address(0)) {
-      bidShares.creator.value = _getDefaultBp();
+      bidShares.creator.value = _getDefaultBp() * (10**16);
     } else {
-      bidShares.creator.value = _getBp(tokenId);
+      bidShares.creator.value = _getBp(tokenId) * (10**16);
     }
     return bidShares;
   }
--- a/contracts/enforcer/HolographRoyalties.sol
+++ b/contracts/enforcer/HolographRoyalties.sol
@@ -408,6 +408,7 @@ contract HolographRoyalties is Admin, Owner, Initializable {
 
   /**
    * @dev Internal function that transfers tokens to all payout recipients.
+   * @dev ERC20 tokens that use fee on transfer are not supported.
    * @param tokenAddress Smart contract address of ERC20 token.
    */
   function _payoutToken(address tokenAddress) private {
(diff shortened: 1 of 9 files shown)


src/enforcer/HolographRoyalties.sol
```diff
@@ -297,18 +297,13 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     address payable[] memory addresses = _getPayoutAddresses();
     uint256[] memory bps = _getPayoutBps();
     uint256 length = addresses.length;
-    // accommodating the 2300 gas stipend
-    // adding 1x for each item in array to accomodate rounding errors
-    uint256 gasCost = (23300 * length) + length;
     uint256 balance = address(this).balance;
-    require(balance - gasCost > 10000, "ROYALTIES: Not enough ETH");
-    balance = balance - gasCost;
+    require(balance > 10000, "ROYALTIES: Not enough ETH");
```
alexanderattar: Should we still keep this check? or at least add a comment why we're limiting payout to be at least 10_000 wei?

src/enforcer/HolographRoyalties.sol
```diff
@@ -324,11 +319,16 @@ contract HolographRoyalties is Admin, Owner, Initializable {
     uint256 balance = erc20.balanceOf(address(this));
     require(balance > 10000, "ROYALTIES: Not enough tokens");
```
alexanderattar: Same thing here

src/enforcer/HolographRoyalties.sol
```diff
@@ -297,18 +297,13 @@ contract HolographRoyalties is Admin, Owner, Initializable {
...
-    // accommodating the 2300 gas stipend
-    // adding 1x for each item in array to accomodate rounding errors
-    uint256 gasCost = (23300 * length) + length;
     uint256 balance = address(this).balance;
-    require(balance - gasCost > 10000, "ROYALTIES: Not enough ETH");
-    balance = balance - gasCost;
+    require(balance > 10000, "ROYALTIES: Not enough ETH");
     uint256 sending;
-    // uint256 sent;
     for (uint256 i = 0; i < length; i++) {
```
alexanderattar, line 323: Multiple wardens suggested we move payments to a pull instead of push approach. That would require more changes and likely keeping a mapping of payees. Before moving forward with that I'd like to get your input Vitto https://github.com/code-423n4/2022-10-holograph-findings/issues/104

src/enforcer/HolographRoyalties.sol
ACC01ADE: We should keep this since if we allow for lower than 10000 WEI, we have a chance that some wallets in payouts won't even get anything due to rounding errors. Plus don't see a point triggering royalties for dust amounts.

src/enforcer/HolographRoyalties.sol
ACC01ADE: Need to remove it from here since some tokens are using no decimals or two decimal places, this would result in $100 for something like USDT getting stuck in account. I think there was a specific C4 issues associated with this one.

src/enforcer/HolographRoyalties.sol
alexanderattar: sounds good

alexanderattar: I removed

src/enforcer/HolographRoyalties.sol
alexanderattar: I removed this one

src/enforcer/HolographRoyalties.sol
ACC01ADE: Our goal with the HolographRoyalties contract is to only collect the royalties and provide a way to access the amounts after. If only one payout recipient is configured, this could be considered as a pull approach. The push approach is set as the default/only option simply due to the fact that we do not plan to add any features/functionality outside of the scope of collecting the royalties. Any additional distribution logic should be handled by an external contract code.

contracts/enforcer/HolographRoyalties.sol
```diff
@@ -696,4 +749,31 @@ contract HolographRoyalties is Admin, Owner, Initializable {
   function getTokenAddress(string memory tokenName) public view returns (address) {
     return _getTokenAddress(tokenName);
   }
+
+  /**
+   * @notice Used to wrap function calls to check if they return without revert regardless of return type.
+   * @dev
```
alexanderattar: @ACC01ADE Would you be able to add an extra note here on the logic. I know you mentioned it while we were chatting yesterday, but wasn't sure the best way to articulate it in writing

src/enforcer/HolographRoyalties.sol
```diff
@@ -597,4 +650,31 @@ contract HolographRoyalties is Admin, Owner, Initializable {
   function getTokenAddress(string memory tokenName) public view returns (address) {
     return _getTokenAddress(tokenName);
   }
+
+  /**
+   * @notice Used to wrap function calls to check if they return without revert regardless of return type.
+   * @dev
+   * @return Returns true if the wrapped function call returns without a revert even if it doesn't return true.
+   */
+  function _callOptionalReturn(
```
ACC01ADE, line 668: In order to accommodate non-compliant ERC-20 tokens (like USDT), which do not return a bool on success, we wrap the ERC20 calls. This allows to not revert when a call returns no data. The logic works by: making a call, checking if return opcode is a revert, if it is then it reverts as well, if it's not then it checks for return data, if return data exists, it is returned as a bool, if return data does not exist (0 length) then success is expected and returns true.

contracts/enforcer/HolographRoyalties.sol
ACC01ADE: Replied in the `src` version of the smart contract. [Here](https://github.com/holographxyz/holograph-protocol/pull/93#discussion_r1038542139)

ACC01ADE:
@alexanderattar I think this is all perfect. One note is that we should add the restriction of maximum array length of 10 for `configurePayouts` in order to avoid users from making configurations that would result in permanent out of gas errors in payout for loops.

sogoiii:
@alexanderattar How are you resolving this issue? https://github.com/code-423n4/2022-10-holograph-findings/issues/451 I see that you removed a bunch of checks.

test/20_royalties_tests.ts
```diff
@@ -540,4 +550,216 @@ describe('HolographRoyalties Contract', async function () {
...
+      const totalRoyalties = ethers.utils.parseEther('1');
+
+      before(async () => {
+        const data = (await royalties.populateTransaction.configurePayouts([owner.address], [10000])).data || '';
+        const tx = await l1.factory.connect(owner).adminCall(royalties.address, data);
+        await tx.wait();
+      });
+
+      it('should be able to withdraw all native token balance', async () => {
+        await owner.sendTransaction({
```
sogoiii, line 567: @alexanderattar I think this should not come from the owner but from somewhere else.

test/20_royalties_tests.ts
```diff
@@ -540,4 +550,216 @@ describe('HolographRoyalties Contract', async function () {
...
+          to: royalties.address,
+          value: totalRoyalties,
+        });
+
+        const accountBalanceBefore = await ethers.provider.getBalance(owner.address);
+        const contractBalanceBefore = await ethers.provider.getBalance(royalties.address);
```
sogoiii, line 573: We should have an expect check for `contractBalanceBefore` that is tied to the `totalRoyalties` balance

test/20_royalties_tests.ts
```diff
@@ -540,4 +550,216 @@ describe('HolographRoyalties Contract', async function () {
...
+
+        const data = (await royalties.populateTransaction.getEthPayout()).data || '';
+        const tx = await l1.factory.connect(owner).adminCall(royalties.address, data);
+        await tx.wait();
+
+        const accountBalanceAfter = await ethers.provider.getBalance(owner.address);
+        const contractBalanceAfter = await ethers.provider.getBalance(royalties.address);
+
+        expect(contractBalanceAfter).to.be.lt(contractBalanceBefore);
```
sogoiii, line 582: We can keep these 2 tests. But I think we should be able to create an expect with a specific value.

test/20_royalties_tests.ts
```diff
@@ -540,4 +550,216 @@ describe('HolographRoyalties Contract', async function () {
...
+    describe('A collection with 1 recipient', async () => {
+      const totalRoyalties = ethers.utils.parseEther('1');
+
+      before(async () => {
+        const data = (await royalties.populateTransaction.configurePayouts([owner.address], [10000])).data || '';
+        const tx = await l1.factory.connect(owner).adminCall(royalties.address, data);
+        await tx.wait();
+      });
+
+      it('should be able to withdraw all native token balance', async () => {
```
sogoiii, line 566: I think we need to add a test for the zero case. You guys mentioned something about this in the chat.

test/20_royalties_tests.ts
```diff
@@ -540,4 +550,216 @@ describe('HolographRoyalties Contract', async function () {
...
+
+        const data = (await royalties.populateTransaction.getTokenPayout(ERC20.address)).data || '';
+        const tx = await l1.factory.connect(owner).adminCall(royalties.address, data);
+        await tx.wait();
+
+        const accountBalanceAfter = await ERC20.balanceOf(owner.address);
+        const contractBalanceAfter = await ERC20.balanceOf(royalties.address);
+
+        expect(contractBalanceAfter).to.be.lt(contractBalanceBefore);
+        expect(accountBalanceAfter).to.be.gt(accountBalanceBefore);
```
sogoiii, line 604: Again, we should be able to check specific values.

test/20_royalties_tests.ts
```diff
@@ -540,4 +550,216 @@ describe('HolographRoyalties Contract', async function () {
...
+
+        const accountABalanceAfter = await ethers.provider.getBalance(anyAddress);
+        const accountBBalanceAfter = await ethers.provider.getBalance(notOwner.address);
+        const contractBalanceAfter = await ethers.provider.getBalance(royalties.address);
+
+        const sixtyPercentOfRoyalties = contractBalanceBefore.sub(contractBalanceAfter).mul(60).div(100);
+        const fortyPercentOfRoyalties = contractBalanceBefore.sub(contractBalanceAfter).mul(40).div(100);
+
+        expect(contractBalanceAfter).to.be.lt(contractBalanceBefore);
+        expect(Number(accountABalanceAfter)).to.be.approximately(
```
sogoiii, line 641: We should add a note to describe why we use approximately here. I think we are using because of the minor differences in computing percentages. But because we define the percentage split above, we should be able to know the exact amount.

test/20_royalties_tests.ts
```diff
@@ -540,4 +550,216 @@ describe('HolographRoyalties Contract', async function () {
...
+        const accountBBalanceAfter = await ethers.provider.getBalance(notOwner.address);
+        const accountCBalanceAfter = await ethers.provider.getBalance(mockAddress);
+        const contractBalanceAfter = await ethers.provider.getBalance(royalties.address);
+
+        const twentyPercentOfRoyalties = contractBalanceBefore.sub(contractBalanceAfter).mul(20).div(100);
+        const fiftyPercentOfRoyalties = contractBalanceBefore.sub(contractBalanceAfter).mul(50).div(100);
+        const thirtyPercentOfRoyalties = contractBalanceBefore.sub(contractBalanceAfter).mul(30).div(100);
+
+        expect(contractBalanceAfter).to.be.lt(contractBalanceBefore);
+        expect(Number(accountABalanceAfter)).to.be.approximately(
```
sogoiii, line 720: We should add a note to describe why we use approximately here. I think we are using because of the minor differences in computing percentages. But because we define the percentage split above, we should be able to know the exact amount.

test/20_royalties_tests.ts
alexanderattar: Added this

test/20_royalties_tests.ts
alexanderattar: where did you have in mind? I think the idea here is it's just funding the contract so there's royalties to pull out. owner is actually `owner = l1.deployer;`. we could change it to be from another account though if it makes more sense

test/20_royalties_tests.ts
alexanderattar: Not going to use approximately anymore now that the contract logic has been fixed

test/20_royalties_tests.ts
alexanderattar: See above

test/20_royalties_tests.ts
alexanderattar: yep updated to use equals instead

test/20_royalties_tests.ts
alexanderattar: Using equals now instead

test/20_royalties_tests.ts
alexanderattar: Done

ACC01ADE:
Made two minor edits/improvements. At this point I believe the PR is ready. We can make any additional tests and final edits part of a new PR (if needed).

ACC01ADE (approved):

PR #99, opened 2022-12-14
Feature/HOLO-685: Final protocol changes for mainnet deployment

ACC01ADE:
## Describe Changes

Making final adjustments for proper mainnet deployment.

- HLG can only be minted on `init`.
- HLG can only be minted once on `init` on **Ethereum** for mainnet.
- Modified tests to still work for localhost environment.
- Switched the slashed tokens to be sent HolographTreasury instead of HLG directly.
- Disable HLG rewards for operator jobs.
- Modified `random` value to be split into two separate values for pod and operator pod selection.
- Modified `random` value to be unique for selection of backup operators.
- Removed temporary code that was in place before `Holographer` had the values applied to it.

### DIFF
--- a/contracts/token/HolographUtilityToken.sol
+++ b/contracts/token/HolographUtilityToken.sol
@@ -132,14 +132,13 @@ contract HolographUtilityToken is ERC20H {
     );
     _setOwner(contractOwner);
     /*
-     * @dev Mint token only if target chain matches current chain.
+     * @dev Mint token only if target chain matches current chain. Or if no target chain has been selected.
      *      Goal of this is to restrict minting on Ethereum only for mainnet deployment.
      */
-    if (block.chainid == targetChain) {
-      /*
-       * @dev Mint only 10 billion tokens on mainnet (10_000_000_000)
-       */
-      HolographERC20Interface(msg.sender).sourceMint(tokenRecipient, tokenAmount * (10**18));
+    if (block.chainid == targetChain || targetChain == 0) {
+      if (tokenAmount > 0) {
+        HolographERC20Interface(msg.sender).sourceMint(tokenRecipient, tokenAmount * (10**18));
+      }
     }
     // run underlying initializer logic
     return _init(initPayload);
--- a/contracts/token/HolographUtilityToken.sol
+++ b/contracts/token/HolographUtilityToken.sol
@@ -126,9 +126,21 @@ contract HolographUtilityToken is ERC20H {
    * @param initPayload abi encoded payload to use for contract initilaization
    */
   function init(bytes memory initPayload) external override returns (bytes4) {
-    address contractOwner = abi.decode(initPayload, (address));
+    (address contractOwner, uint256 tokenAmount, uint256 targetChain, address tokenRecipient) = abi.decode(
+      initPayload,
+      (address, uint256, uint256, address)
+    );
     _setOwner(contractOwner);
-    HolographERC20Interface(msg.sender).sourceMint(contractOwner, 10000000 * (10**18));
+    /*
+     * @dev Mint token only if target chain matches current chain.
+     *      Goal of this is to restring minting on Ethereum only for mainnet deployment.
+     */
+    if (block.chainid == targetChain) {
+      /*
+       * @dev Mint only 10 billion tokens on mainnet (10_000_000_000)
+       */
+      HolographERC20Interface(msg.sender).sourceMint(tokenRecipient, tokenAmount * (10**18));
+    }
     // run underlying initializer logic
     return _init(initPayload);
   }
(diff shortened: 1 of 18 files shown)


contracts/token/HolographUtilityToken.sol
```diff
@@ -126,9 +126,21 @@ contract HolographUtilityToken is ERC20H {
...
-    address contractOwner = abi.decode(initPayload, (address));
+    (address contractOwner, uint256 tokenAmount, uint256 targetChain, address tokenRecipient) = abi.decode(
+      initPayload,
+      (address, uint256, uint256, address)
+    );
     _setOwner(contractOwner);
-    HolographERC20Interface(msg.sender).sourceMint(contractOwner, 10000000 * (10**18));
+    /*
+     * @dev Mint token only if target chain matches current chain.
+     *      Goal of this is to restring minting on Ethereum only for mainnet deployment.
```
alexanderattar: is restring supposed to be `restrict`?

Also how is this enforced if the target chain is passed in via the payload. would it be better to hardcode the target chain to ethereum?

deploy/02_sources.ts
```diff
@@ -224,14 +225,34 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
   );
   hre.deployments.log('the future "HolographRoyalties" address is', futureRoyaltiesAddress);
 
+  const network = networks[hre.networkName];
+
+  let tokenAmount: BigNumber = BigNumber.from('100000000');
+  let targetChain: BigNumber = BigNumber.from(network.chain);
```
alexanderattar: is this so it can run on localhost actually?

deploy/02_sources.ts
```diff
@@ -224,14 +225,34 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
...
     primaryNetwork = networks.localhost;
   } else if (currentNetworkType == NetworkType.testnet) {
+    // one hundred million tokens minted per network on testnets
+    tokenAmount = BigNumber.from('100' + '000' + '000' + '000000000000000000');
     primaryNetwork = networks.ethereumTestnetGoerli;
   } else if (currentNetworkType == NetworkType.mainnet) {
+    // ten billion tokens minted on ethereum on mainnet
+    tokenAmount = BigNumber.from('10' + '000' + '000' + '000' + '000000000000000000');
+    // target chain is restricted to ethereum, to prevent the minting of tokens on other chains
+    targetChain = BigNumber.from(networks.ethereum.chain);
```
alexanderattar, line 249: okay i see it's hardcoded in the deployment instead of the contract

contracts/token/HolographUtilityToken.sol
ACC01ADE: yep it's a typo

deploy/21_hlg.ts
```diff
@@ -65,14 +65,28 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
     process.exit();
   };
 
+  let tokenAmount: BigNumber = BigNumber.from('100000000');
+  let targetChain: BigNumber = BigNumber.from(network.chain);
+  let tokenRecipient: string = deployer.address;
+
   // Future Holograph Utility Token
```
alexanderattar, line 72: Why are the tokens minted in both the `02_sources` script and in this script?

alexanderattar (approved):

deploy/02_sources.ts
ACC01ADE: yep, this is to allow localhost and testnets to work still, while mainnet is strictly restricted to only minting on ethereum.

deploy/21_hlg.ts
ACC01ADE: `02_sources` script is only used to derive future address for configurations, while the `21_hlg` script does the actual deployment.

alexanderattar: aha got it

contracts/token/HolographUtilityToken.sol
ACC01ADE: fixed in next commit

deploy/21_hlg.ts
```diff
@@ -65,14 +65,28 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
     process.exit();
   };
 
+  let tokenAmount: BigNumber = BigNumber.from('100000000');
```
sogoiii, line 68: Lets make this and the same code in `02_sources.ts` a function to protect ourselves from randomly modifying it.

sogoiii (approved):
👍 Looks good to me.

PR #112, opened 2022-12-16
Release/latest testnet deployments 20221216

ACC01ADE:
## Describe Changes

- Updated external deployments from genesis
- Update the logic around when to mint HLG to be more clear and prevent issues between misconfigured mainnet / testnet networks to develop environments
- Fixed a bug in the HolographUtilityToken contract to set the proper number of decimals to match the deploy scripts


deploy/28_factory_deployed_contracts.ts
```diff
@@ -65,9 +65,50 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
       });
       hre.deployments.log('Deployed a "CxipERC721Proxy" empty contract for block explorer verification purposes.');
     }
+
+    const holographUtilityToken: Contract | null = await hre.ethers.getContractOrNull(
```
alexanderattar, line 69: This is just so the HLG token contract shows on block explorers?

deploy/28_factory_deployed_contracts.ts
```diff
@@ -65,9 +65,50 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
...
+        waitConfirmations: 1,
+      });
+      hre.deployments.log('Deployed a "hToken" empty contract for block explorer verification purposes.');
+    }
   }
 };
 
 export default func;
-func.tags = ['Holographer4verify', 'CxipERC721Proxy4verify'];
+func.tags = ['Holographer4verify', 'CxipERC721Proxy4verify', 'HolographUtilityToken4verify', 'hToken4verify'];
```
alexanderattar, line 113: What is this `hToken4verify`?

deploy/28_factory_deployed_contracts.ts
ACC01ADE: Yes, the contract is a dummy and will not function as an actual ERC20 contract but rather just revert

deploy/30_verify.ts
```diff
@@ -24,6 +24,8 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
 
   if (currentNetworkType != NetworkType.local) {
     let contracts: string[] = [
+      'HolographUtilityToken',
```
alexanderattar, line 27: Oh okay I see it gets verified now

alexanderattar (commented):
Just some questions

deploy/28_factory_deployed_contracts.ts
ACC01ADE: Just a tag name to use in case we want to skip rest of deploy scripts and just run deployment for specific contracts that need to be verified.

deploy/30_verify.ts
ACC01ADE: Yep, along with hToken.

deploy/28_factory_deployed_contracts.ts
alexanderattar: I see

deploy/28_factory_deployed_contracts.ts
alexanderattar: where does it revert?

ACC01ADE: It does not revert. We use getContractOrNull purposefully. And then we just check for null.

alexanderattar (approved):
Updates make sense

sogoiii (approved):
👍

PR #115, opened 2022-12-21
Hotfix/incorrect bridge out assembly

ACC01ADE:
## Describe Changes

Fixing incorrectly written logic for assembly in Enforcer bridgeOut requests.

- Fixed bridgeOut assembly code in HolographGeneric
- Fixed bridgeOut assembly code in HolographERC20
- Fixed bridgeOut assembly code in HolographERC721

### DIFF
--- a/src/enforcer/HolographERC20.sol
+++ b/src/enforcer/HolographERC20.sol
@@ -341,17 +341,22 @@ contract HolographERC20 is Admin, Owner, Initializable, NonReentrant, EIP712, Ho
         amount
       );
       assembly {
-        mstore(add(sourcePayload, mload(sourcePayload)), caller())
+        // it is important to add 32 bytes in order to accommodate the first 32 bytes being used for storing length of bytes
+        mstore(add(sourcePayload, add(mload(sourcePayload), 0x20)), caller())
         let result := call(
           gas(),
           sload(_sourceContractSlot),
           callvalue(),
-          sourcePayload,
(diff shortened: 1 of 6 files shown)


src/enforcer/HolographERC20.sol
```diff
@@ -341,17 +341,22 @@ contract HolographERC20 is Admin, Owner, Initializable, NonReentrant, EIP712, Ho
...
       assembly {
-        mstore(add(sourcePayload, mload(sourcePayload)), caller())
+        // it is important to add 32 bytes in order to accommodate the first 32 bytes being used for storing length of bytes
+        mstore(add(sourcePayload, add(mload(sourcePayload), 0x20)), caller())
         let result := call(
           gas(),
           sload(_sourceContractSlot),
           callvalue(),
-          sourcePayload,
-          add(mload(sourcePayload), 32),
```
alexanderattar, line 350: Just wondering if the original intentions here were to skip 32 bytes but it needs to be done in hex via 0x20 instead?

alexanderattar (approved):
The changes generally make sense to me based on the comments. Is there anyway we can test the changes will render the correct token uri data before we deploy and upgrade?

src/enforcer/HolographERC20.sol
ACC01ADE: The use of 32 here was correct. Switched to 0x20 just for clarity and to not confuse bytes with integers.

PR #116, opened 2022-12-22
Improvement/holo 681 update prettier rules for holograph protocol

BravoNatalie:
## Describe Changes

Since [prettier ends all files with a new line by design](https://github.com/prettier/prettier/issues/6360) and there's no option to disable this, the solution here is to add the empty line to the end of all files that are being changed by prettier.

- Changed `abi` task by adding an empty line at the end of the ABI files.
- Created `deploymentsPrettier` task to add a new line at the end of the deployments files.


hardhat.config.ts
```diff
@@ -129,6 +130,46 @@ task('deploy', 'Deploy contracts').setAction(async (args, hre, runSuper) => {
...
+  if (!fs.existsSync('./deployments')) {
+    throw new Error('The directory "deployments" was not found.');
+  }
+
+  function getAllFiles(dirPath: string, arrayOfFiles: string[]) {
+    const files = fs.readdirSync(dirPath);
+
+    arrayOfFiles = arrayOfFiles || [];
+
+    files.forEach(function (file) {
```
alexanderattar: Nit: can this be updated to also use `for (const file of files) {` for consistency

alexanderattar (approved):
Glad to have this annoying thing fixed finally! 👏

sogoiii (changes requested):
@BravoNatalie @alexanderattar Ok i think we have to add this task as part of the "compile" and /or the deploy script in package.json? So we don't have to "remember" to run the command manually.

sogoiii (approved):
👍

PR #120, opened 2022-12-27
Proposal: cross-chain multisig implementation

ACC01ADE:
## Describe Changes

Proposing a cross-chain multisig implementation.

- Leveraging HolographGeneric to create a custom CrossChainMultisig implementation
- Using bridgeIn and bridgeOut to send SecurePackets with multisig-approved data


alexanderattar (commented):
I think I need more context on what the planned use for this is. Maybe it was explained earlier and I missed it or need my memory refreshed after the holiday break. Besides that the changes don't really seem to affect current contracts so I think that should be okay. Let's chat about this when there's time @ACC01ADE

PR #129, opened 2023-01-09
Update references of l1 and l2 to chain1 and chain2 for clarity in test files

alexanderattar:
## Describe Changes

- Simple rename of 12/12 to chain1/chain2 in the tests for clarity since each chain can be either an l1 or l2 as long as it's an EVM chain

### DIFF
--- a/contracts/HolographRegistry.sol
+++ b/contracts/HolographRegistry.sol
@@ -195,10 +195,7 @@ contract HolographRegistry is Admin, Initializable, HolographRegistryInterface {
     assembly {
       contractType := extcodehash(contractAddress)
     }
-    require(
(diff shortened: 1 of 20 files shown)


ACC01ADE:
@alexanderattar Have you checked that there is no reference to `l3`? I remember that at one point when swapping vars, `l3` was used as temp placeholder. Might still be in the code somewhere. Otherwise this is a good update and I approve.

ACC01ADE (approved):

sogoiii (approved):
👍

alexanderattar:
I couldn't find any `l3` referenced in the codebase so I think we're good

PR #133, opened 2023-02-03
Enable multi-chain deployment

ACC01ADE:
## Describe Changes

sample multi-chain deployment


alexanderattar:
We should add a test for this function. Otherwise this seems like a simple an obvious upgrade we should make

sogoiii (approved):
The logic does make sense.

contracts/HolographFactory.sol
```diff
@@ -193,7 +224,7 @@ contract HolographFactory is Admin, Initializable, Holographable, HolographFacto
     DeploymentConfig memory config,
     Verification memory signature,
     address signer
-  ) external {
+  ) public {
```
alexanderattar, line 227: why the change to public visibility?

alexanderattar (approved):

contracts/HolographFactory.sol
sogoiii: I think it was changed because its now being called from within this contract `deployHolographableContractMultiChain`.

PR #142, opened 2023-02-24
Fund Fauce with HLG on updating reference

alexanderattar:
## Describe Changes

- When the HLG reference is updated to a new version we also have to fund the Faucet with the new token

### DIFF
--- a/deploy/29_faucet.ts
+++ b/deploy/29_faucet.ts
@@ -46,6 +46,8 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
 
   if (currentNetworkType == NetworkType.testnet || currentNetworkType == NetworkType.local) {
     if (environment != Environment.mainnet && environment != Environment.testnet) {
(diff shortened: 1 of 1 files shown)


deploy/29_faucet.ts
```diff
@@ -101,6 +103,28 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
         });
         await tx.wait();
         hre.deployments.log('Updated HLG reference');
+        hre.deployments.log('Transferring 1M HLG to faucet');
+        const transferTx = await hlgContract.transfer(
+          futureFaucetAddress,
+          BigNumber.from('1' + '000' + '000' + '000000000000000000'),
+          {
+            ...(await txParams({
```
sogoiii: So for `txParams` and `hre.ethers.provider.estimateGas` I think it would be clearer if those were put into variables and then passed into the `hlgContract.transfer` function

sogoiii (changes requested):

sogoiii (approved):

PR #151, opened 2023-03-29
Feature: final drops contract edits

ACC01ADE:
## Describe Changes

Making the final drops contract edits.

- Reserved namespaces: DropsPriceOracleProxy, EditionsMetadataRendererProxy, and DropsMetadataRendererProxy
- Registering newly reserved namespaces with deployed contracts
- Hardcoding the DropsPriceOracleProxy address into drops contract
- Add name function to the HolographDropERC721 which the metadata renderer requires
- Lock down event config via reverting fallback for HolographDropERC721
- Update drops tests


alexanderattar:
Looks good so far. I'll look out for the hardcoding of the price oracle address into drops contract

ACC01ADE:
At this point I think we need to review this completely.

hardhat.config.ts
```diff
@@ -357,18 +352,20 @@ const config: HardhatUserConfig = {
...
-      avalancheFujiTestnet: AVALANCHE_API_KEY,
-      bscTestnet: BINANCE_SMART_CHAIN_API_KEY,
+      mainnet: process.env.ETHERSCAN_API_KEY || '',
+      goerli: process.env.ETHERSCAN_API_KEY || '',
+      avalanche: process.env.SNOWTRACE_API_KEY || '',
+      avalancheFujiTestnet: process.env.SNOWTRACE_API_KEY || '',
+      polygon: process.env.POLYGONSCAN_API_KEY || '',
+      polygonMumbai: process.env.POLYGONSCAN_API_KEY || '',
+      bsc: process.env.BSCSCAN_API_KEY || '',
+      bscTestnet: process.env.BSCSCAN_API_KEY || '',
```
sogoiii, line 368: Lets not forgot we have to add binance to this section. I know its in the other pr. Just a mental reminder.

sogoiii (approved):
👍

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -755,12 +756,6 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
     emit OpenMintFinalized(msgSender(), config.editionSize);
   }
 
-  function setDropsPriceOracle(address newDropsPriceOracle) external onlyOwner {
```
alexanderattar, line 758: Are we cool with no longer being able to set this?

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -804,4 +794,10 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
       //uint256 id = chainPrepend + uint256(tokenId);
     }
   }
+
+  fallback() external override payable {
+    assembly{
+      revert(0x00, 0x00)
```
alexanderattar: Would it be better to revert with a more descriptive error message like `Function not supported` or `Fallback disabled` or something like that instead of an empty revert?

contracts/drops/token/HolographDropERC721.sol
ACC01ADE: I think it's mandatory to have it hardcoded. Since the drops contract is provided via proxy implementation, we can always upgrade the source with new address, if we need to change it for some reason.

contracts/drops/token/HolographDropERC721.sol
ACC01ADE: Gotta save that gas lol. But ya we could add a simple `revert("Function not found")`.

alexanderattar: Whatever you think is best is cool with me

src/drops/token/HolographDropERC721.sol
```diff
@@ -39,14 +40,19 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
...
-   * @dev bytes32(uint256(keccak256('eip1967.Holograph.osRegistryEnabled')) - 1)
+   * @dev Address of the operator filter registry
    */
-  bytes32 constant _osRegistryEnabledSlot = precomputeslot("eip1967.Holograph.osRegistryEnabled");
+  IOperatorFilterRegistry public constant openseaOperatorFilterRegistry = IOperatorFilterRegistry(0x000000000000AAeB6D7670E522A718067333cd4E);
+
+  /**
+   * @dev Address of the price oracle proxy
+   */
+  IDropsPriceOracle public constant dropsPriceOracle = IDropsPriceOracle(0xA3Db09EEC42BAfF7A50fb8F9aF90A0e035Ef3302);
```
sogoiii, line 55: This is the address to the dropPriceOracle proxy right?

contracts/drops/token/HolographDropERC721.sol
sogoiii: Yea im kinda torn. I do like that it tells me why and i don't have to guess. Maybe it can be super short like `revert("Fn not found")`, `revert("No FN")`, or maybe `revert("¯\_(ツ)_/¯")`

contracts/drops/token/HolographDropERC721.sol
sogoiii: @ACC01ADE @alexanderattar if this is removed where is this being called? are we doing this part of the initializations?

sogoiii (commented):
@ACC01ADE @alexanderattar I think we are forgetting the binance price orcales.

contracts/drops/token/HolographDropERC721.sol
alexanderattar: I updated it to return Function not found via the assembly block

PR #165, opened 2023-05-09
Hotfix/avax drops price oracle updates

ACC01ADE:
## Describe Changes

Hotfixing the issue of not being able to mint $0 drops

- Price oracle now returns 0 wei for 0 input amounts


alexanderattar (changes requested):
This looks good, but I think we need it for all the networks right?

alexanderattar:
> This looks good, but I think we need it for all the networks right?

@ACC01ADE's response:

> The real fix should be applied to the actual Drops contract on each network. All that logic needs to be entirely skipped and save on gas if price is $0

We'll handle the full upgrade once the Spacebrat drop is complete

alexanderattar (approved):

sogoiii (approved):
👍

PR #167, opened 2023-05-22
Holograph Protocol Upgrade for Optimism (Bedrock including)

ACC01ADE:
## Describe Changes

Making necessary changes for supporting Optimism L2 in the Holograph Protocol.

- Optimism L1 Security Fee is not supported and calculated in LayerZeroModule
  - base gas price for L1 fee is set to 100 GWEI for now
- Also configuring minimum gas prices for networks


deploy/10_layer_zero_module.ts
```diff
@@ -109,13 +195,55 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
 
   const holograph = await hre.ethers.getContract('Holograph', deployer);
 
+  const futureOptimismGasPriceOracleAddress = await genesisDeriveFutureAddress(
```
alexanderattar, line 202: This needs to be deployed via Super Cold Storage right?

src/HolographOperator.sol
```diff
@@ -516,7 +516,7 @@ contract HolographOperator is Admin, Initializable, HolographOperatorInterface {
   ) external payable {
     require(msg.sender == _bridge(), "HOLOGRAPH: bridge only call");
     CrossChainMessageInterface messagingModule = _messagingModule();
-    uint256 hlgFee = messagingModule.getHlgFee(toChain, gasLimit, gasPrice, bridgeOutPayload.length);
+    uint256 hlgFee = messagingModule.getHlgFee(toChain, gasLimit, gasPrice, bridgeOutPayload);
```
alexanderattar, line 519: need to upgrade for this change?

src/interface/CrossChainMessageInterface.sol
```diff
@@ -23,6 +23,6 @@ interface CrossChainMessageInterface {
     uint32 toChain,
     uint256 gasLimit,
     uint256 gasPrice,
-    uint256 payloadLength
+    bytes calldata crossChainPayload
```
alexanderattar, line 26: Also just curious why the change?

src/module/LayerZeroModule.sol
```diff
@@ -60,14 +66,16 @@ contract LayerZeroModule is Admin, Initializable, CrossChainMessageInterface, La
       address bridge,
       address interfaces,
       address operator,
+      address optimismGasPriceOracle,
```
alexanderattar, line 69: This is what you mentioned using Uniswap / Sushiswap for right?

deploy/10_layer_zero_module.ts
ACC01ADE: For `minnet` and `testnet` yes, but for `develop` we can deploy this with deployer wallet.

src/module/LayerZeroModule.sol
```diff
@@ -185,6 +193,9 @@ contract LayerZeroModule is Admin, Initializable, CrossChainMessageInterface, La
     require(gasLimit < gasParameters.maxGasLimit, "HOLOGRAPH: gas limit over max");
     (uint256 nativeFee, ) = lz.estimateFees(lzDestChain, address(this), crossChainPayload, false, adapterParams);
     hlgFee = ((gasPrice * gasLimit) * dstPriceRatio) / (10 ** 10);
+    if (toChain == uint32(7) || toChain == uint32(4000000015)) {
```
alexanderattar, line 201: Maybe adding a comment on these chain ids would be helpful so you don't need to look them up if you haven't memorized all the chain ids 😅

src/HolographOperator.sol
ACC01ADE: Yep, will need to upgrade Holograph Operator contract. Otherwise we will not be able to get a precise L1 Security Fee cost. Since zero bytes and non-zero bytes have different costs associated with them.

src/interface/CrossChainMessageInterface.sol
ACC01ADE: To properly calculate the L1 Security Fee, we need entire payload data. Zero bytes and non-zero bytes have different costs associated with them.

src/module/LayerZeroModule.sol
ACC01ADE: No this is for calculating Optimism L1 Security Fee

src/module/LayerZeroModule.sol
alexanderattar: especially for the holographId

ACC01ADE: Good point, made a commit to address this.

src/module/LayerZeroModule.sol
```diff
@@ -330,6 +364,15 @@ contract LayerZeroModule is Admin, Initializable, CrossChainMessageInterface, La
     }
   }
 
+  /**
+   * @dev Internal function used for getting the Holograph Operator Interface
```
alexanderattar: Is this comment a copy/paste that should be updated?

src/module/OVM_GasPriceOracle.sol
```diff
@@ -0,0 +1,160 @@
...
+  function getL1Fee(bytes memory _data) external view returns (uint256) {
+    uint256 l1GasUsed = getL1GasUsed(_data);
+    uint256 l1Fee = l1GasUsed * l1BaseFee;
+    uint256 divisor = 10 ** decimals;
+    uint256 unscaled = l1Fee * scalar;
+    uint256 scaled = unscaled / divisor;
+    return scaled;
+  }
+
+  /**
```
alexanderattar, line 126: This comment is super helpful

src/module/OVM_GasPriceOracle.sol
```diff
@@ -0,0 +1,160 @@
...
+  function getL1GasUsed(bytes memory _data) public view returns (uint256) {
+    uint256 total = 0;
+    for (uint256 i = 0; i < _data.length; i++) {
+      if (_data[i] == 0) {
+        total += 4;
+      } else {
+        total += 16;
+      }
+    }
+    uint256 unsigned = total + overhead;
```
alexanderattar, line 157: Is `overhead` defined somewhere?

alexanderattar: never mind. i just noticed it gets set to storage variable in the constructor

src/interface/CrossChainMessageInterface.sol
alexanderattar: aha good to know

src/module/LayerZeroModule.sol
alexanderattar: 👍

deploy/10_layer_zero_module.ts
alexanderattar: cool that's what i thought

src/module/LayerZeroModule.sol
ACC01ADE: Oops, yep, copy pasta gone bad. Made new commit to fix this.

src/module/OVM_GasPriceOracle.sol
ACC01ADE: Copy pasta from original Optimism Gas Price Oracle contract. Kept the commentary since it explains the logic very very well.

alexanderattar: nice

src/module/LayerZeroModule.sol
alexanderattar: sweet

alexanderattar (approved):
It all generally makes sense to me, but we'll probably need to upgrade develop and let it run for testing before we know if we'll need to make any adjustments

PR #173, opened 2023-05-25
Add tenderly hardhat config and deployment script

alexanderattar:
## Describe Changes

- Adds a new deploy script that verifies the contracts selected on Tenderly which will allow us to get full traces and debug easier

It can be run with:
```
npx hardhat deploy --network polygonTestnet --tags Tenderly
```


deploy/31_tenderly.ts
```diff
@@ -0,0 +1,67 @@
+declare var global: any;
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+import { DeployFunction } from '@holographxyz/hardhat-deploy-holographed/types';
+import { NetworkType, networks } from '@holographxyz/networks';
+import { SuperColdStorageSigner } from 'super-cold-storage-signer';
+
+// ts-ignore because this is not a hardhat plugin
```
alexanderattar: Remove this

hardhat.config.ts
```diff
@@ -374,6 +380,27 @@ const config: HardhatUserConfig = {
...
+  tenderly: {
+    project: 'holograph',
+    username: 'attar',
+    privateVerification: false,
+    forkNetwork: '<fork-chain-id>',
+  },
 };
 
+// Allow hardhat to use short network names
+function mapNetworkKeysByShortKey(networks: Networks) {
```
sogoiii, line 392: @alexanderattar Given that this is used in several spots, should we move this into a file that we export? Would be good to not have duplication.

alexanderattar: it's actually only being used in hardhat.config at the moment

sogoiii: Ah ok, i see. I confused it with `mapFullKeyToShortKey`

sogoiii (approved):
👍 this is a great tool.

PR #179, opened 2023-06-16
Minor protocol improvements

ACC01ADE:
## Describe Changes

- Dropping the use of msg.sender in favor of msgSender() function.
- Improved msgSender() logic to use msg.sender if not holographer call.
- Implemented onAllowance event type in order to remove HolographBridge and HolographOperator overrides for hTokens in the future.
- Adding `sourceExternalCall` in enforcers. This is to allow custom contracts to interact with other non-protocol contracts via holographer.


deploy/31_tenderly.ts
```diff
@@ -12,12 +14,57 @@ function mapFullKeyToShortKey(networks: Networks, fullKey: string) {
...
 
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  /*
   const currentNetworkType: NetworkType = networks[hre.network.name].type;
   if (currentNetworkType == NetworkType.local) {
     hre.deployments.log('Not verifying contracts on localhost networks.');
     return;
   }
 
+  const accounts = await hre.ethers.getSigners();
```
alexanderattar, line 24: was this causing you issues? should we reenable?

ACC01ADE: Yep it was causing issues. Until we can standarize and test this tenderly implementation out, and also add a simple enable/disable toggle, I will keep it disabled.

alexanderattar (approved):

sogoiii (approved):
👍

PR #188, opened 2023-07-26
HOLO-1058: Add protocol mint fee

alexanderattar:
## Describe Changes

- Include a $0.10 Holograph protocol fee for drop mints


contracts/drops/token/HolographDropERC721.sol
```diff
@@ -738,34 +754,20 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
...
-
     // Payout recipient
     (bool successFunds, ) = config.fundsRecipient.call{value: funds, gas: 210_000}("");
     if (!successFunds) {
       revert Withdraw_FundsSendFailure();
     }
 
     // Emit event for indexing
-    emit FundsWithdrawn(sender, config.fundsRecipient, funds, feeRecipient, holographFee);
+    emit FundsWithdrawn(sender, config.fundsRecipient, funds, address(0), 0);
```
alexanderattar: We can update the event or set these to zero. Let me know what you think

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -822,6 +824,21 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
     }
   }
 
+  function _payoutHolographFee(uint256 quantity) internal {
+    // Transfer protocol fee to recipient address
```
alexanderattar: Move this comment

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -822,6 +824,21 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
...
+    uint256 holographFee = getHolographFee(quantity);
+
+    // Payout Holograph fee
+    if (holographFee > 0) {
+      address payable holographFeeRecipient = payable(
```
alexanderattar: Should we add a function to see the recipient?

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -822,6 +824,21 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
...
+        HolographInterface(HolographerInterface(holographer()).getHolograph()).getTreasury()
+      );
+
+      (bool success, ) = holographFeeRecipient.call{value: holographFee, gas: 210_000}("");
```
alexanderattar: Should we revert if success isn't true?

src/drops/token/HolographDropERC721.sol
```diff
@@ -321,6 +324,13 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
       });
   }
 
+  /// @notice The Holograph fee is a flat fee for each mint
+  /// @dev Gets the Holograph protocol fee for amount of mints
+  function getHolographFee(uint256 quantity) public returns (uint256 fee) {
+    holographMintFee = 100000; // $0.10 USD (6 decimals)
```
sogoiii: @alexanderattar @ACC01ADE I was expecting this to call the price oracle to get the mint fee as this will be PER - network.

alexanderattar: It calls it in the payoutHolographFee function

alexanderattar: I would like to move this initialization to another function such as init. Originally I was putting this as a state variable, but was running into issues with this. Would like to discuss during the meeting today

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -420,6 +423,13 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
       });
   }
 
+  /// @notice The Holograph fee is a flat fee for each mint
+  /// @dev Gets the Holograph protocol fee for amount of mints
+  function getHolographFee(uint256 quantity) public returns (uint256 fee) {
+    holographMintFee = 100000; // $0.10 USD (6 decimals)
```
alexanderattar: Move this somewhere else...

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -493,9 +503,11 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
     uint256 quantity
   ) external payable nonReentrant canMintTokens(quantity) onlyPublicSaleActive returns (uint256) {
     uint256 salePrice = _usdToWei(salesConfig.publicSalePrice);
+    uint256 holographMintFeeInWei = _usdToWei(holographMintFee);
 
-    if (msg.value < salePrice * quantity) {
-      revert Purchase_WrongPrice(salesConfig.publicSalePrice * quantity);
+    if (msg.value < (salePrice + holographMintFeeInWei) * quantity) {
+      // This will show the NFT + the Holograph mint fee * quantity in USD
+      revert Purchase_WrongPrice((salesConfig.publicSalePrice + holographMintFee) * quantity);
```
alexanderattar: Update comment:

// The error will display the wrong price in USD

contracts/drops/token/HolographDropERC721.sol
alexanderattar: Updated the event. Seems cleaner

src/drops/token/HolographDropERC721.sol
alexanderattar: Ended up successfully moving this to the init function

contracts/drops/token/HolographDropERC721.sol
alexanderattar: I changed my mind about this. This is something only Holograph can do so it shouldn't be an admin function since that would be the creator. We can update this via upgrades

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -169,6 +169,9 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
    */
   address public marketFilterAddress;
 
+  /// @notice Holograph Mint Fee
+  uint256 public holographMintFee; // $0.10 USD (6 decimal places)
```
alexanderattar: Remove comment

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -509,8 +528,12 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
       revert Purchase_TooManyForAddress();
     }
 
+    // First mint the NFTs
```
alexanderattar, line 532: Should we flip these so that payout happens first @ACC01ADE ?

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -822,6 +832,20 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
     }
   }
 
+  function _payoutHolographFee(uint256 quantity) internal {
+    // Transfer protocol fee to recipient address
+    uint256 holographFee = getHolographFeeUsd(quantity);
```
alexanderattar: Update to use `getHolographFeeWei`

test/forge/HolographDropERC721.t.sol
```diff
@@ -372,34 +365,48 @@ contract HolographDropERC721Test is Test {
...
-    vm.stopPrank();
-    vm.prank(address(0xcafeea3));
-    vm.expectRevert(abi.encodeWithSelector(IHolographDropERC721.OperatorNotAllowed.selector, address(0xcafeea3)));
-    erc721Enforcer.transferFrom(DEFAULT_OWNER_ADDRESS, address(0x666), FIRST_TOKEN_ID);
-    vm.prank(DEFAULT_OWNER_ADDRESS);
-    customSource.manageMarketFilterSubscription(false);
-    vm.prank(address(0xcafeea3));
-    erc721Enforcer.transferFrom(DEFAULT_OWNER_ADDRESS, address(0x666), FIRST_TOKEN_ID);
-  }
+  // TODO: Fix this test. It's failing with the error [FAIL. Reason: ERC721: not approved sender] test_SubscriptionEnabled() (gas: 4685864)
```
alexanderattar, line 370: @ACC01ADE this is unrelated to the fee logic, but I noticed these were failing. I was having some trouble debugging. Can you take a look when you have time, and maybe you'll spot something I'm missing 🙏

test/forge/HolographDropERC721.t.sol
```diff
@@ -420,19 +427,20 @@ contract HolographDropERC721Test is Test {
...
-    bytes memory preBaseCall = abi.encodeWithSelector(
-      IOperatorFilterRegistry.unregister.selector,
-      address(customSource)
-    );
-    customSource.updateMarketFilterSettings(preBaseCall);
-    bytes memory baseCall = abi.encodeWithSelector(IOperatorFilterRegistry.register.selector, address(customSource));
-    customSource.updateMarketFilterSettings(baseCall);
-    vm.stopPrank();
-  }
+  // TODO: Fix this test. I belive it's failing because the default owner is not the owner of the source contract
```
alexanderattar, line 432: Similar to the other failing subscription test

src/drops/token/HolographDropERC721.sol
```diff
@@ -723,6 +731,20 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
...
+
+    // Payout Holograph fee
+    address payable holographFeeRecipient = payable(
+      HolographInterface(HolographerInterface(holographer()).getHolograph()).getTreasury()
+    );
+
+    (bool success, ) = holographFeeRecipient.call{value: nativeFee, gas: 210_000}("");
+    emit MintFeePayout(holographFee, holographFeeRecipient, success);
+  }
+
```
alexanderattar, line 757: Should we emit this in wei or usd?

contracts/drops/token/HolographDropERC721.sol
alexanderattar: I added a new error to revert if fee payout fails

PR #202, opened 2023-09-01
Mainnet into develop

sogoiii:
## Describe Changes

- Syncing develop so all branches are the same.


contracts/HolographOperator.sol
```diff
@@ -496,7 +496,6 @@ contract HolographOperator is Admin, Initializable, HolographOperatorInterface {
    * @notice Receive a cross-chain message
    * @dev This function is restricted for use by Holograph Messaging Module only
    */
-  function crossChainMessage(bytes calldata bridgeInRequestPayload) external payable {
```
alexanderattar: Kind of confused why this is removed in this PR. Should be fixed when we compile again though

alexanderattar (approved):

PR #218, opened 2023-10-24
HolographDropERC721 Upgrades

alexanderattar:
## Describe Changes

- Remove OpenSea operator filter registry because this royalty enforcement mechanism has been deprecated: https://opensea.io/blog/articles/creator-fees-update
- Updates mint fee


src/drops/token/HolographDropERC721.sol
```diff
@@ -65,13 +53,8 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
...
   address public erc721TransferHelper;
 
-  /**
-   * @dev Address of the market filter registry
-   */
-  address public marketFilterAddress;
-
   /// @notice Holograph Mint Fee
-  uint256 public constant HOLOGRAPH_MINT_FEE = 100000; // $0.10 USD (6 decimal places)
+  uint256 public constant HOLOGRAPH_MINT_FEE = 1000000; // $1.00 USD (6 decimal places)
```
ackvf, line 57: So we are increasing the protocol fee for mints from $0.1 to $1? Doesn't it make the fee larger than gas fee on certain networks? Not sure I would like that.

alexanderattar: Yes that's correct. We are raising it to be more aligned with other platforms out there ahead of some big drops and collaborations with partners

feliperodriguess (approved):

alexanderattar:
This was never merged and planned to complete these changes later. There is a new PR for handling this on V2

PR #228, opened 2024-02-07
Upgrade HolographDropERC721

alexanderattar:
## Describe Changes
- Adds HolographDropERC721V2
- Adds a new deployment script for it
- Adds the new contract to the registry
- Update the version to 2 (NOTE: The upgrade is incompatible from previous version therefore making it a MAJOR upgrade)
  - I thought about changing this to a uint32, but because it's a pure function the optimization would be minimal.
- Remove OpenSea operator filter registry because this royalty enforcement mechanism has been deprecated: https://opensea.io/blog/articles/creator-fees-update
  - Updates the deployment scripts to be compatible with the above
- Updates the price oracle proxy reference to the latest address `0xeA7f4C52cbD4CF1036CdCa8B16AcA11f5b09cF6E`
- Moves the `holographMintFee` fee into the treasury (initially set to $1.00)
- Adds functions for getting and setting the mint fee
- Adds tests to check the new functions

To run tests:

1. run `yarn ganache-x2` or `yarn anvil`
2. `npx hardhat deploy --network localhost`
3. `forge test`

image


contracts/drops/token/HolographDropERC721.sol
```diff
@@ -117,12 +122,6 @@ import {SalesConfiguration} from "../struct/SalesConfiguration.sol";
...
 
-import {IMetadataRenderer} from "../interface/IMetadataRenderer.sol";
-import {IOperatorFilterRegistry} from "../interface/IOperatorFilterRegistry.sol";
-import {IHolographDropERC721} from "../interface/IHolographDropERC721.sol";
-
-import {IDropsPriceOracle} from "../interface/IDropsPriceOracle.sol";
-
 /**
  * @dev This contract subscribes to the following HolographERC721 events:
  *       - beforeSafeTransfer
```
sogoiii: @alexanderattar i think we can remove `beforeTransfer` and `beforeSafeTransfer` from here because those are being removed in this pr.

contracts/drops/token/HolographDropERC721.sol
```diff
@@ -834,10 +678,21 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
...
     }
   }
 
+  function _getHolographMintFee() public view returns (uint256) {
+    address payable treasuryProxyAddress = payable(
+      HolographInterface(HolographerInterface(holographer()).getHolograph()).getTreasury()
+    );
+
+    HolographTreasury treasury = HolographTreasury(treasuryProxyAddress);
+    return treasury.holographMintFee();
```
sogoiii: @alexanderattar I think you are accessing the storage value right? is that more efficient than going through the method `getHolographMintFee` ?

alexanderattar: yes i'm updating this

contracts/drops/token/HolographDropERC721.sol
alexanderattar: Good catch. Although I think that requires us to recompute the event config bytes.

alexanderattar: Updated the event config

sogoiii (approved):
👍

BravoNatalie (approved):

PR #245, opened 2024-03-11
Deploy testnet faucet and other improvements

alexanderattar:
## Describe Changes

- Deploy Faucet on testnet env
- Making logging more explicit in deployment scripts
- Standardize logging to use `console.log` (i.e. no more mixed console.log and `hre.deployments.log`)
- Make sure 10 billion HLG is sent regardless of env / network type to keep everything consistent
- Add extra warning for mainnet
- Throw error if `primaryNetwork` isn't undefined (it shouldn't.. this just a paranoid precaution)
- Add missing `await` at the transfer call to send 1 million HLG to the operator
- Improves logging in interfaces script to actually log the interface that's getting checked
  - BEFORE
  image
  - AFTER
  image


deploy/21_hlg.ts
```diff
@@ -58,21 +58,23 @@ const func: DeployFunction = async function (hre1: HardhatRuntimeEnvironment) {
...
+  let primaryNetwork: Network | undefined; // Initialize primaryNetwork variable
   if (currentNetworkType == NetworkType.local) {
-    // one billion tokens minted per network on local testing
-    tokenAmount = BigNumber.from('1' + '000' + '000' + '000' + '000000000000000000');
+    // ten billion tokens minted per network on local testnet
+    tokenAmount = BigNumber.from('10' + '000' + '000' + '000' + '000000000000000000');
     primaryNetwork = networks.localhost;
   } else if (currentNetworkType == NetworkType.testnet) {
-    // one hundred million tokens minted per network on testnets
-    tokenAmount = BigNumber.from('100' + '000' + '000' + '000000000000000000');
```
alexanderattar, line 68: This 100 million line was getting overwritten on testnet anyways as can be seen by the amount in the testnet multisig

image

sogoiii (approved):
👍 I like the console log change.

PR #253, opened 2024-03-27
Update beam terminology to bridge

alexanderattar:
## Describe Changes

- Update `beam` to `bridge`


alexanderattar:
Even though we can update the code in the repo without updating the bytecode, we can't reverify with the updated code and I'd rather keep the repo in sync with block explorers so I'm closing this

PR #259, opened 2024-04-04
Feature/precomput hashes script

alexanderattar:
## Describe Changes

- Add a pre-commit command that runs the precompute slot script
- Make it so the precompute slot script updates any slots that are found


Yashiru (approved):
Nice 👌, it works very well

README.md
```diff
@@ -293,15 +293,29 @@ future "Holograph" address is 0x17253175f447ca4B560a87a3F39591DFC7A021e3
...
+
+A note on contract slots. The protocol uses the EIP1967 standard heavily for addressing storage location via slots. We use a consistent format across the codebase
+in the form of:
+
+eip1967.<contract-name>.<storage-variable-name>
+
+Additionally we have a utility script that indentify this within the contracts directory as a pre-commit and auto-hash it into it's proper format prior to pushing updates to the repo. An example of how to add a new storage variable to a contract via a hashed slot is as followed:
+
+1. Add a variable to your contracts storage such as `bytes32 constant _interfacesSlot = precomputeslot("eip1967.Holograph.interfaces");`.
+2. On commit, the pre-commit script will run and update the contents of the `precompute` slot function into the output of the hash that is actually stored in the contract.
```
sogoiii: @alexanderattar Its great that this works via the git commit hook. But I think that the hook is the safety net, not the primary dev work flow. I think we should describe how the developer should use the script by calling it manually. Because I would imagine we are adding a slot, and updating a test somewhere. its weird to require doing a commit to force the conversion. 

## Maybe something like this

1. Add a variable to your contracts storage such as `bytes32 constant _interfacesSlot = precomputeslot("eip1967.Holograph.interfaces");`. Add slots below existing slots. Never add them in between existing slots. 
2. Run the `npx ts-node scripts/precompute-solidity-hashes.ts` script to update the contents of the `precompute` slot function into the output of the hash that is actually stored in the contract.
3. Annotate the converted line with a comment so the original value is captured. See existing slot to see how we comment these values. We should be able to see what the string version of the hash is.
4. These constants provide access to important variables at a unique location in a consistent manner.

Note, that if you do not manually call the `precompute-solidity-hashes.ts` script, we have a git hook that will automatically convert it.

sogoiii (changes requested):

sogoiii (approved):
👍

PR #261, opened 2024-04-05
improved hTokenBalance and extractNativeToken scripts

Yashiru:
## Changes

I've improved hTokenBalance and extractNativeToken scripts to make them easier to use and to debug.

- Refactored them into the script directory.
- Replaced the `--contract` parameter with `--token` to allow human-readable values ([hEth, hMatic, hAvax...]).
- Added transaction debug data in the extractNativeToken script.
- Added a user confirmation requirement before transaction execution in the extractNativeToken script.

> [!Important]
> I was unable to add `getHTokenAddress` in `scripts/utils/helper.ts` due to redundant Hardhat imports.

> [!Tip]
> We can install [colors](https://www.npmjs.com/package/colors) to avoid using raw colors in logs. However, I preferred not to add a new js dependency.

 - [x] I have performed a self-review of my code.
 - [x] Code styles have been enforced.
 - [ ] All Hardhat tests are passing.


scripts/utils/addresses.ts
```diff
@@ -0,0 +1,32 @@
+import { NetworkType } from '@holographxyz/networks';
+
+type TokenSymbol = 'hETH' | 'hBNB' | 'hAVAX' | 'hMATIC' | 'hMNT';
+type HTokenAddresses = {
+  [network in NetworkType]: {
+    [token in TokenSymbol]?: string;
+  };
+};
+
+export const tokenAddresses: HTokenAddresses = {
```
alexanderattar: Since this is moved to a file called addresses.ts, I think it makes sense to update the variable to be specific since there could be other contract addresses in this file. Maybe `hTokenAddresses`

scripts/utils/addresses.ts
```diff
@@ -0,0 +1,32 @@
+import { NetworkType } from '@holographxyz/networks';
+
+type TokenSymbol = 'hETH' | 'hBNB' | 'hAVAX' | 'hMATIC' | 'hMNT';
+type HTokenAddresses = {
```
alexanderattar: Maybe update this to `HTokenAddressMap` to not conflict with the variable

alexanderattar (approved):

scripts/h-token-balance.ts
```diff
@@ -0,0 +1,50 @@
...
+import { HardhatRuntimeEnvironment } from 'hardhat/types';
+import { TokenSymbol, getHTokenAddress } from './utils/addresses';
+
+/**
+ * Task to get the hToken Balance of an account
+ * @param contract The address of the hToken contract
+ * @param account The address of the account
+ *
+ * Run this task with:
+ * npx hardhat hTokenBalance --token [hETh, hMatic, hAvax...] --account [accountAddress] --network [the target network]
```
sogoiii: change --account to --recipient like the extract native token script.

sogoiii (approved):
👍

PR #265, opened 2024-04-09
Move contracts to src

alexanderattar:
## Describe Changes

We are transitioning to Foundry which uses `src` as the default contracts directory so in an effort to be consistent with that project layout

image


sogoiii (changes requested):
@alexanderattar How did you end up testing this? 

1. the package.json has an array called `files`. Do we have to remove `contract` and add `src`?
2. In the README we callout the fact that we have the `src` and `contract` folder. I think we can remove all that stuff now.

Yashiru (commented):
I think we can change `contract/` to `src/` in `deploy/30_verify.ts` L104
```solidity
const contractFullName = `contracts/drops/oracle/${contract}.sol:${contract}`;
```
⬇️
```solidity
const contractFullName = `src/drops/oracle/${contract}.sol:${contract}`;
```

alexanderattar:
I have made all the suggested changes. I am going to merge this because so I can cleanly create a new branch off of it

PR #268, opened 2024-04-11
Feature/holo 1480 lazy mint

Yashiru:
## Describe Changes

- Add lazy mint feature (onlyOwner function to set the encrypted data for each batch)
- Add `canLazyMint` but we need to define if it is usefull or not and if yes => what mechanic to include in it
- Add tokenURI feature 
- Started to work on the test structure (Fixtures and constants)
- Started coding a first test to test the complete workflow in order to confirm it before writing it into the notion (at the very beginning, 3 lines of code)
- Add contractURI functionality


src/HolographGenesis.sol
```diff
@@ -246,11 +246,7 @@ contract HolographGenesis {
    * @param approve Approval status in the message
    * @return bytes32 Ethereum signed message hash
    */
-  function getMessageHash(
```
alexanderattar, line 249: Uh oh @Yashiru I think we need to sync our prettier settings. These files shouldn't have this code updated like this

src/token/CustomERC721.sol
```diff
@@ -430,13 +456,14 @@ contract CustomERC721 is NonReentrant, ERC721H, BatchMintMetadata, DelayedReveal
...
    * @dev Mints multiple editions to the given list of addresses.
+   * @dev TODO: Double chFeck if we need to use arrays for encryptedBaseUris and dataArray
    * @param recipients list of addresses to send the newly minted editions to
    */
   function adminMintAirdrop(
     address[] calldata recipients
   ) external onlyOwner canMintTokens(recipients.length) returns (uint256) {
     unchecked {
-      for (uint256 i = 0; i < recipients.length; i++) {
+      for (uint256 i = 0; i != recipients.length; i++) {
```
alexanderattar, line 473: Is this just a gas optimization?

src/token/CustomERC721.sol
```diff
@@ -23,14 +23,17 @@ import {SalesConfiguration} from "../drops/struct/SalesConfiguration.sol";
...
 
 /**
  * @dev This contract subscribes to the following HolographERC721 events:
  *       - customContractURI
  *
  *       Do not enable or subscribe to any other events unless you modified the source code for them.
  */
-contract CustomERC721 is NonReentrant, ERC721H, BatchMintMetadata, DelayedReveal, ICustomERC721 {
+contract CustomERC721 is NonReentrant, ERC721H, LazyMint, DelayedReveal, ICustomERC721 {
+  using Strings for uint256;
```
alexanderattar, line 36: Do we need this library?

src/extension/LazyMint.sol
```diff
@@ -0,0 +1,56 @@
...
+ *  minting a non-zero balance of NFTs of those tokenIds.
+ */
+
+abstract contract LazyMint is ILazyMint, BatchMintMetadata {
+  /// @dev The sender is not authorized to perform the action
+  error LazyMintUnauthorized();
+  error LazyMintInvalidAmount();
+
+  /// @notice The tokenId assigned to the next new NFT to be lazy minted.
+  uint256 internal nextTokenIdToLazyMint;
```
alexanderattar, line 21: Okay so under the hood this keeps track of the sequence of tokens that have the next encrypted tokenURI. 

i.e. `nextTokenIdToLazyMint` - `amount`

so for example the next tokenId is 101 and the amount is 100, the next batch would be from 101-201

src/token/CustomERC721.sol
```diff
@@ -432,7 +432,7 @@ contract CustomERC721 is NonReentrant, ERC721H, LazyMint, DelayedReveal, ICustom
     uint256 _amount,
     string calldata _baseURIForTokens,
     bytes calldata _data
-  ) public override onlyOwner returns (uint256 batchId) {
+  ) public override returns (uint256 batchId) {
```
alexanderattar, line 442: We should not remove `onlyOwner` from this. It needs to only be callable by the deployer of the contract

test/foundry/utils/Constants.sol
```diff
@@ -67,4 +67,37 @@ library Constants {
   function getDropsEventConfig() internal pure returns (uint256) {
     return 0x0000000000000000000000000000000000000000000000000000000000040000;
   }
+
```
alexanderattar, line 70: Seeing that this is growing, I think it would be a good idea to create a new directory under `foundry` called `constants` and then create a new file for contract specific constants. We can decide on a naming convention, but maybe something like `CustomERC721.Constants.sol`. `Core.Constants.sol`, etc

src/token/CustomERC721.sol
Yashiru: yes exactly

src/token/CustomERC721.sol
Yashiru: not anymore, I used it before my last commit

src/extension/LazyMint.sol
Yashiru: yes exaclty

src/token/CustomERC721.sol
Yashiru: Yes of course, It is mandatory, I don't know why it is removed

Yashiru: Oh I know !

I just copy/past the version from thirdweb exemple to confirm that my function was the same, but their version has not the `onlyOwner` 👍🏻

src/HolographGenesis.sol
Yashiru: yep you are right

Maybe we can use forge, it will probably not change any contract bytecode

src/token/CustomERC721.sol
alexanderattar: cool cool

src/token/CustomERC721.sol
alexanderattar: okay cool got it

alexanderattar (approved):

alexanderattar (approved):

PR #270, opened 2024-04-12
Hardhat to foundry test rewrite

fgregoribattista:
## Describe Changes

The idea of this PR is to rewrite the tests from hardhat to foundry

- 01_deployment_tests_l1
  - Suite result: ok. 18 passed; 0 failed; 15 skipped; finished in 59.97ms (146.93ms CPU time)
 - 02_deployement_tests_l2
   - Suite result: ok. 19 passed; 0 failed; 3 skipped; finished in 43.22ms (46.56ms CPU time)

In both test 01 and 02 we have contracts without validating the bytecode, because there are a couple of contracts that we need to be able to get their addresses and these change in each deploy run. So they will be fixed in conjunction with the foundry deploy script.

- 03_erc20_enforcer_tests_l1
  - Suite result: ok. 76 passed; 0 failed; 8 skipped; finished in 36.08ms (67.78ms CPU time)

A validation test of the erc20 interface has been incorporated, in case the erc20 holograph interface changes tomorrow and the test is modified. In this case, by keeping the validation of an erc20, we ensure that the interface is **always supported.** 
Also there are several tests that are skipped in hardhat and would not be working, they were written but also skipped.
On the other hand there is a section in hardhat that only describes the test. These tests do not exist in hardhat and were not created in foundry.

- Also updated the foundry version to the latest stable version.


sogoiii (approved):
👍

PR #276, opened 2024-04-23
Countdown rework

Yashiru:
## Describe Changes

I've reworked the countdown so that it is consumed by time, not just a representative limit of a maximum static supply.
Now, when a purchase is made, the `currentTokenId + amount` must be less than the `currentMaxSupply()`.

### Current max supply
return the maximum supply according to the current `block.timestamp`. 
It returns the initial max supply before the start date minus the `elapsedTime / mintInterval`
=> elapsed time is the time elapsed since start date


src/token/CustomERC721.sol
```diff
@@ -212,27 +227,28 @@ contract CustomERC721 is NonReentrant, ContractMetadata, InitializableLazyMint,
...
 
+    // Set the lazy mint initialized status to true to prevent this function from being called again
     _setLazyMintInitialized();
   }
 
-  /**
-   * PUBLIC NON STATE CHANGING FUNCTIONS
-   * static
-   */
+  /* -------------------------------------------------------------------------- */
```
alexanderattar, line 248: Nice i like these updates

alexanderattar (approved):
Nice work. This has the new mechanic and tests pass as expected

PR #277, opened 2024-04-23
Fix compile from hardhat to foundry

lucas-marc:
## Describe Changes

- Changing compile task from hardhat to foundry, to avoid having multiple artifacts for the same contracts
- Added a new SKIP_NATSPEC_SMELLS env variable to avoid the pre_commit natspec smells check, as there are a lot of natspecs missing
- Fixed natspecs smells check, as it was looking for the contracts folder, instead of the src one.
- Updates the lock file
- Temporary disabled git add . in pre-commit


sogoiii (approved):
👍

PR #282, opened 2024-04-25
Add minter role and modifier

alexanderattar:
## Describe Changes

- Add minter role and modifier
- Add `mintTo` function that can only be called by `minter`
- Update deployment scripts
- Small update to test fixture

### DIFF
--- a/src/interface/ICustomERC721.sol
+++ b/src/interface/ICustomERC721.sol
@@ -107,9 +107,8 @@ interface ICustomERC721 {
   /// @notice Getter for the mint interval
   function MINT_INTERVAL() external view returns (uint256);
 
-  // TODO: Do we need this?
-  // /// @notice Getter for the minter role
-  // function minter() external view returns (address);
+  /// @notice Getter for the minter role
+  function minter() external view returns (address);
 
   /// @notice Admin function to update the sales configuration settings
   /// @param publicSalePrice public sale price in ether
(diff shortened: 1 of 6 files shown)


src/interface/ICustomERC721.sol
```diff
@@ -106,6 +107,10 @@ interface ICustomERC721 {
   /// @notice Getter for the mint interval
   function MINT_INTERVAL() external view returns (uint256);
 
+  // TODO: Do we need this?
```
alexanderattar, line 110: what do you guys think about this?

src/interface/ICustomERC721.sol
```diff
@@ -106,6 +107,10 @@ interface ICustomERC721 {
...
+  // /// @notice Getter for the minter role
+  // function minter() external view returns (address);
+
   /// @notice Admin function to update the sales configuration settings
```
alexanderattar, line 113: I'm not sure if we need this in the interface and what visibility we should give it. Solidity will automatically create a getter for public storage vars, but maybe to be consistent we add a function. What do you think @Yashiru ? I mostly just want to stay consistent

Yashiru: I always include explicit function definitions in the interface for all callable methods. This ensures that other devs can interact with our contract through `ICustomERC721(address).anyViewFunction()` without needing to import the full `CustomERC721` contract in their projects.

Yashiru (commented):
We can remove the `adminMint` function in this PR I think

src/interface/ICustomERC721.sol
alexanderattar: okay awesome. very good point

PR #294, opened 2024-05-01
Feature/holo 1480 fix get owner

Yashiru:
## Describe Changes

⚠️ DON'T MERGE


alexanderattar:
This was just a temporary solution to help get tests passing, but needs to be handled more appropriately a different way

PR #295, opened 2024-05-01
Fix HolographOperator security issue

ACC01ADE:
**Removing static messaging module address, which is related to Holograph V1 and should not have access to Holograph V2 under any circumstances.**

## Describe Changes

- Removed static reference to `0x777C19834a1A2FF6353a1E9cfb7C799ed7943a11` which is a `LayerZeroModuleProxy` contract from **Holograph V1** and should not have access to **Holograph V2**.

### DIFF
--- a/src/HolographOperator.sol
+++ b/src/HolographOperator.sol
@@ -497,10 +497,7 @@ contract HolographOperator is Admin, Initializable, HolographOperatorInterface {
    * @dev This function is restricted for use by Holograph Messaging Module only
    */
   function crossChainMessage(bytes calldata bridgeInRequestPayload) external payable {
-    require(
-      msg.sender == address(_messagingModule()) || msg.sender == 0x777C19834a1A2FF6353a1E9cfb7C799ed7943a11,
-      "HOLOGRAPH: messaging only call"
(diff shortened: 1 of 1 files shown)


src/HolographOperator.sol
```diff
@@ -497,10 +497,7 @@ contract HolographOperator is Admin, Initializable, HolographOperatorInterface {
    * @dev This function is restricted for use by Holograph Messaging Module only
    */
   function crossChainMessage(bytes calldata bridgeInRequestPayload) external payable {
-    require(
-      msg.sender == address(_messagingModule()) || msg.sender == 0x777C19834a1A2FF6353a1E9cfb7C799ed7943a11,
```
alexanderattar, line 501: We had identified this and will deploy to remove the V1 module. Thanks for opening the PR @ACC01ADE

alexanderattar (approved):

src/HolographOperator.sol
ACC01ADE: I see the fix is being applied via the `feature/holograph-operator-update` branch. Closing this PR.

PR #299, opened 2024-05-03
Remove price oracle

Yashiru:
## Describe Changes

I removed the price oracle feature, allowing the contract to handle native tokens.


alexanderattar (approved):
This looks good, but there is a known issue where what seems to be the exact same tokenURI is not matching in the Foundry tests. The base64 encoded data returns the exact same JSON though. Not sure if a bug in Foundry or something else. Let's try to figure that out

PR #304, opened 2024-05-10
Feature/holo 1480 ownership s1

Yashiru:
## Describe Changes

I changed the ownership management in the HolographERC721 and the CountdownERC721.

- Give ownership to the source contract in the init of the HolographERC721 enforcer
- Call setOwner in the init of the source contract to set the owner to the one defined in the init payload
- Override the function (and modifiers) that get the owner in the source contract to make them call the `owner()` or `getOwner` function of the enforcer contract


src/token/CountdownERC721.sol
```diff
@@ -265,7 +274,15 @@ contract CountdownERC721 is NonReentrant, ERC721H, ICountdownERC721 {
...
     setStatus(1);
 
-    return _init(initPayload);
+    _init(initPayload);
+
+    (bool success, ) = msg.sender.call(
+      abi.encodeWithSignature("setOwner(address)", initializer.initialOwner)
+    );
+
+    // require(success, "HOLOGRAPH: initOwner failed");
```
alexanderattar: I think we need to figure out why this isn't returning true for success. even if it still is setting the owner that seems a little concerning

Yashiru: Actually working on that 👍🏻

Yashiru: The following commit fixed that
https://github.com/holographxyz/holograph-protocol/pull/304/commits/627064390834dcfe4e0aff8d0df8b58003dbf360

alexanderattar (changes requested):
I think we will not merge this until we also have the script to fix previously deployed contracts. At the moment though we are also blocked on merging this until the tests are passing successfully @Yashiru so please try to focus on resolving those

Yashiru:
All tests are passing now 🎉 

Capture d’écran 2024-05-16 à 14 18 35

src/interface/IHolographERC721Errors.sol
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: UNLICENSED
```
alexanderattar: This can be changed to `MIT`

src/interface/IHolographERC721Errors.sol
```diff
@@ -0,0 +1,134 @@
...
+/*
```
alexanderattar: This can be removed, all new contracts no longer need this license header and we'll just use `MIT` going forward

src/drops/token/HolographDropERC721.sol
```diff
@@ -318,6 +319,17 @@ contract HolographDropERC721 is NonReentrant, ERC721H, IHolographDropERC721 {
     setStatus(1);
 
     return _init(initPayload);
+
```
alexanderattar: yes we do for new drops

alexanderattar: this should be for `HolographDropERC721V2.sol` only though

PR #305, opened 2024-05-14
Check the decoded token uri

Yashiru:
## Describe Changes

I created a new test for the tokenURI base64 decoding feature. 
This test checks if the base64 data decoding are equals to the final json expected to make sure that even if the last bytes are padded, the final JSON is valid anyway.


alexanderattar (approved):
Nice looks good

PR #313, opened 2024-05-17
First pass at a script to decode safe transactions

alexanderattar:
## Describe Changes

- This is an experiment for decoding safe transactions programatically


alexanderattar:
This did not end up working because txData can be variable length so it is non-trivial and not possible to iterate over nested calls without preknown data inputs / lengths

PR #316, opened 2024-05-23
HOLO-1535: Enable LayerZero V2 Compatibility

alexanderattar:
## Describe Changes

The updates the message module to work with the new LayerZero PriceFeed since the logic was moved and the previous logic in our LayerZeroModule is currently broken on many chains LZ has migrated over to V2. This primarily affects many testnet chains including all that were switched from Goerli to Sepolia, along with all new mainnets that LZ has added support for post the V2 launch.

- Add IWorker.sol interface
- Add IPriceFeed.sol interface
- Add LayerZeroModuleV2.sol contract
- Add proxy for the above
- Add new deploy scripts to deploy and attach HolographOperator to the new message module


src/abstract/OApp.sol
```diff
@@ -0,0 +1,138 @@
+// SPDX-License-Identifier: UNLICENSED
```
alexanderattar: @Yashiru You can replace this with just `// SPDX-License-Identifier: MIT` for all new contracts added to the repo. The patent-pending stuff is all deprecated for anything new

PR #318, opened 2024-05-24
HOLO-1536 save init payload (WIP)

Yashiru:
## Describe Changes

I updated all the ERC721 sources contracts to make them store the payload used in their initialization.

For each source contracts
- Add INIT_PAYLOAD private storage variable
- Update init function to store the init payload in the INIT_PAYLOAD
- Add a getter function that decode the INIT_PAYLOAD

> [!NOTE]
> The initPayload could be directly storaged ass a struct type for each contracts, but in the `CountdownERC721` and the `CustomERC721` contracts, build failed with a stack too deep error in the constructors. So to create a "standard" in all contracts, I opted for a private storage variable of type `bytes` and a getter function to decode the bytes. **This also made it possible to have a getter function named according to each contract.**

### DIFF
--- a/test/foundry/fixtures/HolographERC721Fixture.t.sol
+++ b/test/foundry/fixtures/HolographERC721Fixture.t.sol
@@ -1,0 +1,218 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.13;
+
+import {Test, Vm} from "forge-std/Test.sol";
+
+import {HolographTreasury} from "src/HolographTreasury.sol";
+import {DummyDropsPriceOracle} from "src/drops/oracle/DummyDropsPriceOracle.sol";
+import {CountdownERC721Initializer} from "src/struct/CountdownERC721Initializer.sol";
+import {DeploymentConfig} from "src/struct/DeploymentConfig.sol";
+import {CustomERC721SalesConfiguration} from "src/struct/CustomERC721SalesConfiguration.sol";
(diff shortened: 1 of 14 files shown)


test/foundry/fixtures/HolographERC721Fixture.t.sol
```diff
@@ -0,0 +1,218 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.13;
```
alexanderattar, line 2: Nice. Thanks for adding this fixture. I think it'll be great if we can make all the tests relatively consistent

Yashiru: Yes you are right, I'd like to improve/refactor the entire foundry test base to standardize everything and make it consistent.

But we'll have to wait until everything has been migrated from hardhat to foundry I think.

wdyt ?

alexanderattar: yeah probably. we can do what we can on the tests we're writing but we probably shouldn't touch what the Think and Dev team are done until they reach a good stopping point

alexanderattar (approved):
Nice. This looks good!

BravoNatalie (approved):

sogoiii (changes requested):
Can this also be applied to the ERC20 contracts?

PR #353, opened 2024-08-07
HAL-07 Incompatibility with non-standard ERC20

Yashiru:
## Audit report ticket
https://www.halborn.com/portal/reports/protocol#8.7

## Describe Changes
- Updated all tokenUtility transfer to use the SafeErc20 OZ library


alexanderattar (approved):
Changes look good to me. We would need to validate that this doesn't break anything on testnet, which might be the more time consuming task at hand

sogoiii (approved):
👍

